PR #10, opened 2020-03-16
Unity compatibility patch / .NET 2 + 2 Subset

AdamFrisby:
This set of patches allow Backtrace to be compiled for Unity in .NET 2/.NET 2 Subset mode. We have also added #ifdefs for compatibility with versions prior to Unity 2017.4. Patches were built against the latest release version (2.0.4).

Key changes:
- Removed nameof(...) operator, and swapped to strings.
- Changed NET35 to (NET35 || NET_2_0 || NET_2_0_SUBSET) to match corresponding Unity auto-defines. (sorry did this with search/replace, it's a little messy -- too many to do manually in any reasonable time period)
- Converted $"" to String.Format
- Removed method?. and replaced with if/then/else or ternary operator

Special notes have been added with "// TODO" comments for sections which should be checked and tested more robustly.

Mostly untested.


konraddysput:
Hey @AdamFrisby 


Thanks for all these changes. I will check your changes and make sure that they're working correctly for supported Unity verisons/runtimes! 

I will keep you posted with all my progress.

src/Model/BacktraceData.cs
```diff
@@ -146,8 +146,8 @@ public static BacktraceData Deserialize(string json)
         {
             var @object = BacktraceJObject.Parse(json);
 
-            var classfiers = @object["classifiers"]?
-                .Select(n => n.Value<string>()).ToArray() ?? null;
+            var classfiers = @object == null ? null : @object["classifiers"]
+                                 .Select(n => n.Value<string>()).ToArray() ?? null;
```
konraddysput: `?? null`  - I think its not compatible with .NET 2.0.  I will prepare fix for this change. We don't need to use `??` here.

src/Model/BacktraceStackFrame.cs
```diff
@@ -123,15 +123,15 @@ public BacktraceStackFrame(StackFrame frame, bool generatedByException)
...
+            Library = string.IsNullOrEmpty(SourceCodeFullPath) ? (frame.GetMethod() != null ? frame.GetMethod().DeclaringType.ToString() : null) : SourceCodeFullPath;
 
             SourceCode = generatedByException
                     ? Guid.NewGuid().ToString()
                     : string.Empty;
             Column = frame.GetFileColumnNumber();
             try
             {
-                MemberInfo = frame.GetMethod()?.MetadataToken;
+                MemberInfo = (frame.GetMethod() != null ? frame.GetMethod().MetadataToken : default(int)); // TODO: Check me
```
konraddysput: We can use null value here.

konraddysput:
Thanks @AdamFrisby for these changes! We will prepare new alpha version with these changes to unblock you! However we will continue testing these plugin changes with our testing team. Thanks!

PR #11, opened 2020-04-03
Added flags for Create menu

jeromemaurey:
Added flags for ability to generate configuration files for Configuration, Client Configuration, Database Configuration.
Unity_u33cyclk5H

### DIFF
--- a/src/Model/BacktraceClientConfiguration.cs
+++ b/src/Model/BacktraceClientConfiguration.cs
@@ -5,6 +5,7 @@ using UnityEngine;
 namespace Backtrace.Unity.Model
 {
     [Serializable]
+    [CreateAssetMenu(fileName = "DefaultBacktraceClientConfiguration", menuName = "Backtrace/Client Configuration", order = 2)]
     public class BacktraceClientConfiguration : ScriptableObject
     {
         public string ServerUrl;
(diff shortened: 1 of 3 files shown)


konraddysput:
hi @jasoncdavis0 ,

Thanks for your changes. Could you please give me more details why would you like to initialize database object without Backtrace client?

thanks!

jeromemaurey:
Hello,

This is mostly UX, following the flow from packages like HDRP.
Unity_NEj8Cpheii

Cheers,

J.

sbahra:
Hi Jerome! Thanks for the patch! During UX testing, we found that exposing three different configurations as such tended to make it easier for users to shoot themselves in the foot. All the three items are already configurable through the existing window which couples things correctly, while still allowing advanced power users to decouple for more advanced use-cases.

@jeromemaurey Alright if we close this out or was there a specific need or problem that's unaddressed?

jeromemaurey:
Closed it.

I still think the main config should be exposed to the Create menu. Let me know if you want me to submit a PR for that file only ;)

sbahra:
Ok! Sorry, I missed this message. @konraddysput What do you think?

konraddysput:
@sbahra - latest Unity plugin version expose Backtrace configuration under Create menu. I think we add this sugestion to our source code

PR #17, opened 2020-04-27
WebGL compatibility

vlussenburg:
DO NOT MERGE

Problem: there's no access to System.Net.* in WebGL [source](https://docs.unity3d.com/Manual/webgl-networking.html?_ga=2.99307034.562744094.1588025670-1704455353.1584134619)

Solution: Add special #preprocessor for WEBGL. Risk: might be the case for more platforms, requiring more cases to be added.

Problem 2: GUID is generated on every report. That's fine (although wasteful/superfluous) as long as it's a deterministic algorithm. For those cases where no unique `guid` can be generated, that means each and every report will have a new `guid`.

Solution 2: By statically generating the `guid`, this problem is circumvented within one 'game session'. Tested on WebGL Safari & Chrome.

Problem 3: this problem is not covered by a test. If someone unwitting removes the exception check (it looks hokey), the problem will be reintroduced

Solution 3: IN PROGRESS - how can we test these compiler flags in an unit test? Not really right? Now my test process is a (long) cycle of compiling and manually testing in a browser = suck @konraddysput . We could do this instead: `Application.platform == RuntimePlatform.WebGLPlayer` and override that in the unit test. Blah. Probably not that big of a deal.

DO NOT MERGE


vlussenburg:
Tested on Unity Editor 2018.4 on Mac.
WebGL on Chrome and Safari

konraddysput:
Hey @vlussenburg 

Let me help you with those problems:

1. I think the best way to solve it is just to add a preprocessor directive with a newly generated GUID. I've validated backtrace-js code and there we're always returning new GUID as an UUID value in JSON object.

2. Guid will always be unique. There is a small chance that a new guid won't be unique. However to track "user session" we would need to store new guid in local storage, which might be tricky because our user have to be aware of this. I think it's fine to store new generated guid in local storage and then generate new if local storage is empty - as long as we let users know about that.

3. @vlussenburg in unit tests you can choose target platform - in this can webgl. If you will switch platform in unit tests, Unity will run code available under preprocessor directive.

vlussenburg:
Hey @konraddysput 

I think I explained some things wrong 😆 

1. Agreed. And that is what the code in the PR currently does. What I arbitrarily elected to do is to keep that logic as close as possible to the retrieval of the NIC MAC - which is impossible on WebGL. If you see where I put the preprocessor you see I made the change as small as possible. If look at the commit before that, you see I had the preprocessor in the parent function which I thought was a bit more messy, but really who cares :)

2. I mistyped (it was not about the `guid` being unique). What I mean - and check the `UnityDemo` project on `cd03` for examples for this - that with the code on master, if I disable the MAC retrieval (which blows up), I still that every error that my game throws gets a unique `guid`. This is because the `generateMachineId()` function is called for every report. By making this function `static` and calling it only once, at least now we'll have a static `guid` for the game session. Indeed, if the user restarts the game, a new `guid` is generated. If you google around you can indeed find examples of a `guid` being persisted in `UserPrefs` which is an option, but I would treat that as a separate feature request.

3. Cool. I'm gonna ask you on Slack how to do this.

vlussenburg:
- Offline database doesn't work (would have to the be implemented using `PlayerPrefs`).

Issue #28, opened 2020-07-02
Race condition on _backtraceLogManager


vlussenburg:
In `BacktraceClient.cs`:

```cs
            // set maximum game object depth
            _gameObjectDepth = Configuration.GameObjectDepth == 0
                ? 16 // default maximum game object size
                : Configuration.GameObjectDepth;
            CaptureUnityMessages();
            _reportLimitWatcher = new ReportLimitWatcher(Convert.ToUInt32(Configuration.ReportPerMin));

            _nativeClient = NativeClientFactory.GetNativeClient(Configuration, name);
#if UNITY_2018_4_OR_NEWER

            BacktraceApi = new BacktraceApi(
                credentials: new BacktraceCredentials(Configuration.GetValidServerUrl()),
                ignoreSslValidation: Configuration.IgnoreSslValidation);
#else
            BacktraceApi = new BacktraceApi(new BacktraceCredentials(Configuration.GetValidServerUrl()));
#endif
            _backtraceLogManager = new BacktraceLogManager(Configuration.NumberOfLogs);
```

The effect of `CaptureUnityMessages();` (callback being registered) requires `_backtraceLogManager` to be set.

Fix: make sure `_backtraceLogManager` is initialized before Unity callback is registered.

vlussenburg:
Symptom: these messages in your log:

```
07-03 02:03:06.610 20172 20213 E Unity   : NullReferenceException: Object reference not set to an instance of an object
07-03 02:03:06.610 20172 20213 E Unity   :   at Backtrace.Unity.BacktraceClient.HandleUnityMessage (System.String message, System.String stackTrace, UnityEngine.LogType type) [0x00009] in <d0f5068404a941f186af1a8ab5607519>:0 
07-03 02:03:06.610 20172 20213 E Unity   :   at UnityEngine.Application.CallLogCallback (System.String logString, System.String stackTrace, UnityEngine.LogType type, System.Boolean invokedOnMainThread) [0x00014] in <d71499eab75645c19df09ba17af9a506>:0 
07-03 02:03:06.610 20172 20213 E Unity   : UnityEngine.DebugLogHandler:Internal_Log(LogType, String, Object)
07-03 02:03:06.610 20172 20213 E Unity   : UnityEngine.DebugLogHandler:LogFormat(LogType, Object, String, Object[])
07-03 02:03:06.610 20172 20213 E Unity   : UnityEngine.Logger:Log(LogType, Object)
[...]

PR #31, opened 2020-07-07
Log manager execution order

konraddysput:
Previously, when the developer edited `BacktraceClient` source code and added `Debug.Log` (or any other log method) between `BacktraceLogManager` initialization and `CaptureUnityMessages` method, in the worst case this will cause an exception in `BacktraceClient` source code.

This diff move `BacktraceLogManager` initialization to `CaptureUnityMessages` method. We decided to initialize manager there because `BacktraceLogManager` is a required dependency of `CaptureUnityMessages` method.

https://github.com/backtrace-labs/backtrace-unity/issues/28

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Backtrace Unity Release Notes
 
+## Version 3.0.1
+- `BacktraceLogManager` (class responsible for storing log data) will be initialized in `CaptureUnityMessages` method to prevent situation, when custom log from `BacktraceClient` can throw an unhandled exception.
+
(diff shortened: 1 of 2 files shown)


konraddysput:
Personally I love IoC frameworks, however adding more dependency to our plugin will increase plugin size (critical for mobile).

On the other hand, Backtrace Client right now needs to initialize a lot of stuff on startup and on refresh. Because of that it's better to read 'story' that is represented by methods, than whole code. Maybe it's worth to change method name from CaptureUnityMessages to something else?

Let's rethink initialization refractor on another diff

Issue #26, opened 2020-07-01
Generated screenshot on Desktop build are way too large


vlussenburg:
Plugin uses `byte[] bytes = tex.EncodeToPNG();` to encode a lossless screenshot, resulting in a 2.1MB file on my laptop. Suggest to use `byte[] bytes = tex.EncodeToJPG();`

vlussenburg:
![4595383d-6580-45e5-a65c-ff851ffe3f40](https://user-images.githubusercontent.com/726645/86296251-9a466200-bbb5-11ea-8fc2-b50edc227083.png)

vlussenburg:
![8645f453-4c18-43e3-b579-ec61ab480a4c](https://user-images.githubusercontent.com/726645/86296315-bfd36b80-bbb5-11ea-9024-4c41e59f7a84.jpg)
The JPG is about 170kB

vlussenburg:
Implemented in the current (sloppy) fork:
https://github.com/vlussenburg/backtrace-unity

vlussenburg:
- [ ] Windows Standalone
- [x] MacOS Editor
- [ ] MacOs Standalone
- [x] iOS
- [x] Android

konraddysput:
It's kinda weird because Unity tools is generating .png file as well. I will compare image size today and I will bring more info later.

konraddysput:
Png file on Windows:

![32455586-ffbf-4ba2-a94a-12beebc00abd](https://user-images.githubusercontent.com/9386268/86766281-9ea5cb80-c04a-11ea-80a0-5766ef7d313f.png)


Size: 376KB

Jpg on Windows:
![a7c69c1a-99d1-4599-933b-b768c03647b0](https://user-images.githubusercontent.com/9386268/86766504-ee849280-c04a-11ea-9bd8-2b0e258da597.jpg)

Size: 31,6KB

konraddysput:
Please see Backtrace-Unity library changes in the diff available here: https://github.com/backtrace-labs/backtrace-unity/pull/32

konraddysput:
Fix will be available in version 3.0.1 (already on master)

vlussenburg:
Tested master @ 9f298881bb070a7da8ac748a0fe701b45becc4c8

- [x]  Windows Standalone
- [x]  MacOS Editor
- [ ]  MacOs Standalone
- [x]  iOS
- [x]  Android
- [x]  WebGL

PR #32, opened 2020-07-07
Store crash image in JPG format

konraddysput:
Previously, the Backtrace-Unity library tried to generate a crash image in PNG format. Because of that, the image size was unacceptable. Comparison between JPG and PNG showed that JPG size is much smaller (the same image - PNG: 376KB vs JPG: 31,6KB).

This diff changes the default image type to JPG.

https://github.com/backtrace-labs/backtrace-unity/issues/26


vlussenburg:
LGTM

PR #33, opened 2020-07-07
Database improvements

konraddysput:
This diff allows our users to dynamically create a path to BacktraceDatabase based on `Application` object properties (`dataPath` or `persistentDataPath`). 

In addition to that this diff fixes a problem with database creation order. Right now, BacktraceDatabase will create the database directory before directory validation function (that will prevent from starting Database)

Testing strategy:
* Diff has unit tests that validate generated path to the database based on short database path (for example './database' will create database path: "Application.dataPath/database" instead) and that tests interpolated strings.


vlussenburg:
Hey! I would personally drop the interpolated feature unless you know a specific customer use case. `dataPath` is supposed to be read only: [here](https://docs.unity3d.com/ScriptReference/Application-dataPath.html).

If you're doing it for backwards compatibility I'd clearly mark it as deprecated to be dropped in 4.x.

konraddysput:
Few users already asked for this feature and it's available on our backlog. Without this feature you won't be able to use Unity Application properties. 

This change doesn't break public API. We didnt introduce any breaking change here. We don't need to store dead code in our code base.

konraddysput:
According to `dataPath` suggestion:

```
Like the documentation of dataPath shows it simply returns the path to your "game". This depends on the platform you run on.. But as a few examples:

in the editor it returns the path to the Assets folder inside your project folder.

in a standalone build it returns the path to the "data" folder where the game data is packed into Unity's custom asset format.

in an Android build it points to the APK file of your app.

The usage of dataPath has to be differenciated between "editor use" and "runtime use". In the editor you would use this path only in editor scripts. Runtime scripts should keep in mind what this path is used for in the build version. The Assets folder doesn't exist in a built game. At runtime you might want to use that path to manually load / access resources that you manually ship along with your build.
```

dataPath might be used in dev environment - our users might have different configuration files for dev and productions. In my case it's much easier to manage and watch database in the unity directories than somewhere on the hard drive.

Few of our users wanted to store data in persistentDataPath - this is mostly a case for production use. 

We're supporting now both.

konraddysput:
Waiting for @jasoncdavis0 changelog/readme suggestions until we will merge those changes to master.

Issue #27, opened 2020-07-02
[Android 8 and lower] ANR crash detection broken + breaks entire plugin when enabled


vlussenburg:
```07-02 16:18:22.840 19267 19279 E Unity   : FileNotFoundException: Could not find file "/Path to not existing file"
07-02 16:18:22.840 19267 19279 E Unity   :   at System.IO.FileStream..ctor (System.String path, System.IO.FileMode mode, System.IO.FileAccess access, System.IO.FileShare share, System.Int32 bufferSize, System.Boolean anonymous, System.IO.FileOptions options) [0x001b7] in <2fa7a6a452ca43df998f07fd1486c0df>:0 
07-02 16:18:22.840 19267 19279 E Unity   :   at System.IO.FileStream..ctor (System.String path, System.IO.FileMode mode, System.IO.FileAccess access, System.IO.FileShare share) [0x00000] in <2fa7a6a452ca43df998f07fd1486c0df>:0 
07-02 16:18:22.840 19267 19279 E Unity   :   at (wrapper remoting-invoke-with-check) System.IO.FileStream..ctor(string,System.IO.FileMode,System.IO.FileAccess,System.IO.FileShare)
07-02 16:18:22.840 19267 19279 E Unity   :   at System.IO.File.OpenRead (System.String path) [0x00000] in <2fa7a6a452ca43df998f07fd1486c0df>:0 
07-02 16:18:22.840 19267 19279 E Unity   :   at System.IO.File.ReadAllBytes (System.String path) [0x00000] in <2fa7a6a452ca43df998f07fd1486c0df>:0 
07-02 16:18:22.840 19267 19279 E Unity   :   at PlayerShip.Fire () [0x000c1] in <96b16c5b85c14628918aa833cc8a7a29>:0 
07-02 16:18:22.840 19267 19279 E Unity   :   at PlayerShip.Update () [0x0005f] in <96b16c5b85c14628918aa833cc8a7a29>:0 
07-02 16:18:22.840 19267 19279 E Unity   :  
07-02 16:18:22.840 19267 19279 E Unity   : (F
07-02 16:18:22.920 19267 19279 E Unity   : NullReferenceException: Object reference not set to an instance of an object
[...]

PR #34, opened 2020-07-08
ANR watchdog initialization

konraddysput:
Previously Backtrace native client initialized ANR watchdog via a static java method. This caused a problems in Android 7 `AndroidJavaException: java.lang.NoSuchMethodError: no non-static method "Lbacktrace/io/backtrace_unity_android_plugin/BacktraceANRWatchdog;.<init>()V"`

This diff changes a way how we initialize ANR watchdog - right now we will use BacktraceANRWatchdog constructor instead.


konraddysput:
Those changes should fix this issue: 

https://github.com/backtrace-labs/backtrace-unity/issues/27

PR #41, opened 2020-08-14
did not work on Unity 2018.4

vlussenburg:
@konraddysput could you see if my changes still work on your end, assuming you're on a newer editor?


konraddysput:
Thanks @vlussenburg for this pull request. Actually .meta files are pretty important for compilation for specific android platforms. If you will take a look into meta file definition you will see that .meta file contains platform specific condition - when compilation should include assemblies and when it shouldn't. 

We need to adjust .meta files to older versions of Unity. I will take a look into this. However right now I think it's better to close this review.

PR #42, opened 2020-08-18
Update SymbolsUpload.cs

vlussenburg:


konraddysput:
Thanks Vincent! I tested your changes on Windows and it looks like it's working correctly. 

Vincent, could you please provide an URL to the Backtrace project where you uploaded your data? I would like to confirm that symbols in `symbols.zip` archive looks correct in Backtrace from MAC.

In addition to that - Did you change line endings in this diff? It looks like Github detected few weird changes - on the top of the SymbolsUpload file and in the end.

I think we're ready to merge those changes just after we check how symbols look like in Backtrace. 

Thanks!

konraddysput (approved):

PR #54, opened 2020-12-03
Version 3.2.6

konraddysput:
## Version 3.2.6
- `BacktraceClient` will apply sampling only to unhandled exceptions type error.
- Fixed annotations nullable value,
- Renamed `BacktraceUnhandledException` Classifier to `UnhandledException`


konraddysput:
wrong diff range

PR #55, opened 2020-12-03
Version 3.2.6 - Sampling improvements

konraddysput:
## Version 3.2.6
- `BacktraceClient` will apply sampling only to unhandled exceptions type error.
- Fixed annotations nullable value,
- Renamed `BacktraceUnhandledException` Classifier to `UnhandledException`


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Backtrace Unity Release Notes
 
+## Version 3.2.6
+- `BacktraceClient` will apply sampling only to unhandled exceptions type error.
+- Fixed annotations nullable value,
+- Renamed `BacktraceUnhandledException` Classifier to `UnhandledException`
```
sbahra, line 6: - "classifier" not "Classifier"
- Let's just call it `error` and add another classifier for the log level.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Backtrace Unity Release Notes
 
+## Version 3.2.6
+- `BacktraceClient` will apply sampling only to unhandled exceptions type error.
+- Fixed annotations nullable value,
```
sbahra, line 5: erroneous comma

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Backtrace Unity Release Notes
 
+## Version 3.2.6
+- `BacktraceClient` will apply sampling only to unhandled exceptions type error.
```
sbahra, line 4: `to errors lacking exception information.`

Runtime/BacktraceClient.cs
```diff
@@ -20,7 +20,7 @@ public class BacktraceClient : MonoBehaviour, IBacktraceClient
     {
         public BacktraceConfiguration Configuration;
 
-        public const string VERSION = "3.2.5";
+        public const string VERSION = "3.2.6";
```
sbahra, line 23: FWIW: We should've done 3.3.X instead following semver.org

Runtime/Model/BacktraceUnhandledException.cs
```diff
@@ -365,7 +365,7 @@ private BacktraceStackFrame SetDefaultStackTraceInformation(string frameString)
         /// </summary>
         private void TrySetClassifier()
         {
-            Classifier = "BacktraceUnhandledException";
+            Classifier = "UnhandledException";
```
sbahra, line 368: `error` `warning`

Runtime/Model/JsonData/Annotations.cs
```diff
@@ -40,7 +40,12 @@ public Annotations(Exception exception, int gameObjectDepth)
             var environmentVariables = new Dictionary<string, string>();
             foreach (DictionaryEntry variable in Environment.GetEnvironmentVariables())
             {
-                environmentVariables.Add(variable.Key.ToString(), Regex.Escape(variable.Value.ToString() ?? "NULL"));
+                if (variable.Key == null)
```
sbahra, line 43: I don't see this bug fix in release notes. Also, how does one disable environment variables?

sbahra (changes requested):

Runtime/BacktraceClient.cs
konraddysput: ah yes - but now we should continue with 3.2.x I guess?

Runtime/Model/JsonData/Annotations.cs
konraddysput: > I don't see this bug fix in release notes.
Adding info to changelog - I focused there only on information that impacts users.
>  Also, how does one disable environment variables?
Info is available here:
https://github.com/backtrace-labs/backtrace-unity#data-privacy

CHANGELOG.md
konraddysput: done

PR #63, opened 2021-03-02
Update README.md

vlussenburg:


README.md
```diff
@@ -192,13 +195,21 @@ The backtrace-unity library includes support for capturing Android NDK crashes a
...
-## ANRs and Hangs
+## ANRs and Hangs <a name="anr-reporting"></a>
+
+When configuring the backtrace-unity client for an Android deployment, programmers will have a toggle available in the backtrace-unity GUI in the Unity Editor to enable or disable ANR or Hang reports. This will use the default of 5 seconds. This will be made configurable in a future release. 
+
+The `error.type` for these reports will be `Hang`.
+
+## (Early Access) Low Memory Reports <a name="oom-reporting"></a>
+
+Backtrace can detect low memory situations for a game running in Unity on Android devices, and attempt to generate an error report with an associated dump object for further investigation. When configuring the backtrace-unity client for an Android deployment, programmers will have a toggle available in the backtrace-unity GUI in the Unity Editor to enable or disable sending Low memory warnings to Backtrace.
```
konraddysput, line 206: ` When configuring the backtrace-unity client for an Android deployment,` this feature is available also on iOS.

on Android when we receive a low memory warning, our native android library will set an attribute memory.warning to true and memory.warning.date to current date (with time). With that being said, when crash occurred because of our of memory - for example bad allocaiton, user can recognize it by checking those two attributes. If memory.warning is true and memory.warning.date is close to dump creation date then we're sure that this is a dump generated by our native integration


on iOS - we're monitoring application state and reason why application exited - by comparing for example application.verison, system.version etc. When system.version, application.version are the same, debugger wasn't connected and low memory warning information are available then we're reporting low memory warning.

README.md
```diff
@@ -192,13 +195,21 @@ The backtrace-unity library includes support for capturing Android NDK crashes a
...
+
+## (Early Access) Low Memory Reports <a name="oom-reporting"></a>
+
+Backtrace can detect low memory situations for a game running in Unity on Android devices, and attempt to generate an error report with an associated dump object for further investigation. When configuring the backtrace-unity client for an Android deployment, programmers will have a toggle available in the backtrace-unity GUI in the Unity Editor to enable or disable sending Low memory warnings to Backtrace. 
 
-When configuring the backtrace-unity client for an Android deployment, programmers will have a toggle available in the backtrace-unity GUI in the Unity Editor to enable or disable ANR or Hang reports. This will use the default of 5 seconds. The `error.type` for these reports will be `Hang`.
+The `error.type` for these reports will be `Low Memory`.
 
-## Low Memory Reports (Early Access)
+Our integration is built in such a way that if multiple low memory warnings are being generated in rapid succession, we will only send a report every 2 minutes. Currently, this is not configurable.
```
konraddysput, line 210: we're not sending report every 2 minutes. We're regenerating memory warning data in application state cache. So when applicaiton crashed because of oom exception on the next applicaiton session we will have the latest available version of memory warning.

konraddysput (changes requested):
What do you think about my suggested changes?

README.md
jasoncdavis0: Thanks for specifying the attributes we set (memory.warning and memory.warning.date) in Android. 

Not sure if text adds anything meaningful for iOS, so I wouldn't update anything there unless i'm missing something.

jasoncdavis0 (approved):
Approved with Konrad's suggested addition of iOS support and the attributes we set for Android.

README.md
vlussenburg: Sorry @konraddysput I missed this. I don't understand what the 2 minutes timer does then. Can you elaborate?

konraddysput: Mobile games can generate a lot low memory warnings. In this case on iOS when we receive a memory warning, we should be smarter on generating low memory report.  So when the game received a new memory warning, but we already stored memory warning in the application cache, we don't need to generate it once again. 2 minutes means that we will generate a new application state when previous memory warning is older than 2 minutes.

PR #64, opened 2021-03-04
Optimize screenshot size and upload speed, allow for error logs to be filtered client side

oddgames-david:
Reasons for the change

1. Screenshots are currently sitting between 500kb - 800kb on high end phones, this is going to quickly fill up storage.
2. Scaling screenshots on GPU and sending a smaller size file is likely faster even if JPG compression takes slightly longer due to upload time being significantly lower.
3. Using Unity's ScreepCapture API to ensure highest compatibility with UI
4. Sometimes you want to be able to stop Error logs from being reported as they aren't always important in all projects


oddgames-david:
This doesn't have backwards compatibility for 2017.4

oddgames-david:
I added some defines to use methods that are available in 2017.4 but I don't have time to test it on that version so if you are interested, feel free to grab it. 

https://github.com/oddgames/backtrace-unity/

PR #66, opened 2021-03-09
Detect false-positive ANRs report and prevent ANR watchdog from sending them

konraddysput:
# Why

When the user closed an application just after the ANR watchdog ANR check, the next time when an application goes from background to foreground, the ANR watchdog might report a false-positive ANR. To prevent this situation we need to also detect when an application goes to background/foreground and prevent/allow ANR checks depends on the application state.

# Technical background
* Game objects update method works only when an application is in the foreground, however, our ANR watchdog background thread is able to compare main/background thread time even when an application goes to background. In this situation when an application goes to the background just after ANR check and before any update method, we will report an ANR situation. When an application goes to foreground/background the `nativeClient` instance sets `_preventAnr` flag to false/true.
* This patch also contains few improvements 
  - we need to be sure that ANR watchdog uses the most recent time available in the main thread - because of that we changed Update to LateUpdate
  - to avoid ThreadAbortExecption we started comparing thread state instead of just aborting thread - long story short, this answer contains all needed resources to read about it: https://stackoverflow.com/a/19048316/4818730
  - to prevent reporting ANR when an application goes from background to thread, the nativeClient instance set lastUpdatedTime to 0 - which algorithm will treat like a starting point for detecting ANRs.

# Testing strategy
**Move application to background on Android/iOS just after ANR detection method and wait more than 5 seconds**
- previously this use case report an ANR
- now it doesn't 
- tested only manually on Android device/iOS simulator

No unit tests.

# Commercialization
This diff will be merged to branch release/3.3.4. After merging all technical changes we will work on the documentation/changelog improvements.

### DIFF
--- a/Runtime/Native/Android/NativeClient.cs
+++ b/Runtime/Native/Android/NativeClient.cs
@@ -20,9 +20,9 @@ namespace Backtrace.Unity.Runtime.Native.Android
 
         /// <summary>
         /// Determine if the ANR background thread should be disabled or not 
-        /// for some time of period.
+        /// for some period of time.
         /// This option will be used by the native client implementation
-        /// once applicaiton goes to background/foreground
+        /// once application goes to background/foreground
         /// </summary>
         volatile internal bool _preventAnr = false;
 
@@ -348,10 +348,10 @@ namespace Backtrace.Unity.Runtime.Native.Android
 
                         lastUpdatedCache = _lastUpdateTime;
                     }
-                    // make sure when ANR happened just after going to foreground
-                    // we won't false positive ANR report
                     else if (lastUpdatedCache != 0)
                     {
+                        // make sure when ANR happened just after going to foreground
+                        // we won't false positive ANR report
                         lastUpdatedCache = 0;
                     }
                     Thread.Sleep(5000);
@@ -417,7 +417,7 @@ namespace Backtrace.Unity.Runtime.Native.Android
         /// <summary>
         /// Pause ANR detection
         /// </summary>
-        /// <param name="stopAnr">True - if native client should pause an ANR detection</param>
+        /// <param name="stopAnr">True - if native client should pause ANR detection"</param>
         public void PauseAnrThread(bool stopAnr)
         {
             _preventAnr = stopAnr;
--- a/Runtime/Native/Android/NativeClient.cs
+++ b/Runtime/Native/Android/NativeClient.cs
@@ -16,7 +16,20 @@ namespace Backtrace.Unity.Runtime.Native.Android
     internal class NativeClient : INativeClient
     {
         // Last Backtrace client update time 
-        internal float _lastUpdateTime;
+        volatile internal float _lastUpdateTime;
+
+        /// <summary>
+        /// Determine if the ANR background thread should be disabled or not 
+        /// for some time of period.
+        /// This option will be used by the native client implementation
+        /// once applicaiton goes to background/foreground
+        /// </summary>
+        volatile internal bool _preventAnr = false;
+
+        /// <summary>
+        /// Determine if ANR thread should exit
+        /// </summary>
+        volatile internal bool _stopAnr = false;
 
         private Thread _anrThread;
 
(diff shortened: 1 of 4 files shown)


Runtime/Native/Android/NativeClient.cs
```diff
@@ -16,7 +16,20 @@ namespace Backtrace.Unity.Runtime.Native.Android
...
     {
         // Last Backtrace client update time 
-        internal float _lastUpdateTime;
+        volatile internal float _lastUpdateTime;
+
+        /// <summary>
+        /// Determine if the ANR background thread should be disabled or not 
+        /// for some time of period.
+        /// This option will be used by the native client implementation
+        /// once applicaiton goes to background/foreground
```
sbahra, line 25: Typo here.

Runtime/Native/INativeClient.cs
```diff
@@ -40,5 +40,11 @@ internal interface INativeClient
         /// Disable native integration
         /// </summary>
         void Disable();
+
+        /// <summary>
+        /// Pause ANR thread
```
sbahra, line 45: Is it worth exposing "Thread" here? Seems like an implementation detail.

Runtime/Native/iOS/NativeClient.cs
```diff
@@ -16,7 +16,20 @@ namespace Backtrace.Unity.Runtime.Native.iOS
...
     {
         // Last Backtrace client update time 
-        internal float _lastUpdateTime;
+        volatile internal float _lastUpdateTime;
+
+        /// <summary>
+        /// Determine if the ANR background thread should be disabled or not 
+        /// for some time of period.
+        /// This option will be used by the native client implementation
+        /// once applicaiton goes to background/foreground
```
sbahra, line 25: Typo

Runtime/Native/iOS/NativeClient.cs
```diff
@@ -16,7 +16,20 @@ namespace Backtrace.Unity.Runtime.Native.iOS
     internal class NativeClient : INativeClient
     {
         // Last Backtrace client update time 
-        internal float _lastUpdateTime;
+        volatile internal float _lastUpdateTime;
+
+        /// <summary>
+        /// Determine if the ANR background thread should be disabled or not 
+        /// for some time of period.
```
sbahra, line 23: period of time

Runtime/Native/iOS/NativeClient.cs
```diff
@@ -139,35 +152,45 @@ public void HandleAnr(string gameObjectName, string callbackName)
...
+                        else
+                        {
+                            reported = false;
+                        }
+
+
+                        lastUpdatedCache = _lastUpdateTime;
                     }
-                    else
+                    // make sure when ANR happened just after going to foreground
```
sbahra, line 183: These comments blocks are a bit hard to follow. I would suggest putting braces on same line or put the comment inside the block.

Runtime/Native/iOS/NativeClient.cs
```diff
@@ -228,9 +251,18 @@ public void Disable()
...
             {
-                _anrThread.Abort();
+                _stopAnr = true;
             }
         }
+
+        /// <summary>
+        /// Pause ANR detection
+        /// </summary>
+        /// <param name="stopAnr">True - if native client should pause an ANR detection</param>
```
sbahra, line 261: "should pause ANR detection"

sbahra (changes requested):

Runtime/Native/INativeClient.cs
konraddysput: I'm against it - I don't see a use case of it and can only generate more issue that we think. Right now we're managing it and to do that right, user need to have an access to nativeClient implementation which is internal right now.

Runtime/Native/iOS/NativeClient.cs
konraddysput: done - thanks

Runtime/Native/iOS/NativeClient.cs
konraddysput: done - thanks.

Runtime/Native/iOS/NativeClient.cs
konraddysput: done - thanks.

Runtime/Native/iOS/NativeClient.cs
konraddysput: done - thanks

rqbacktrace (approved):
As you mentioned previously, we might want to see if we can unify Android/iOS ANR algorithm used by Unity here. Otherwise looks good to me!

sbahra (approved):

vlussenburg (changes requested):
This requires a (unit) test.
a) to prove the problem exists (before the code change)
b) to verify the problem is fixed
and of course c) to show all scenarios we support.

konraddysput:
> This requires a (unit) test.
> a) to prove the problem exists (before the code change)
> b) to verify the problem is fixed
> and of course c) to show all scenarios we support.

I wish Unity can provide an easy tool to mock moving device to background and watch background thread time to synchronize them somehow - you were able to reproduce it by just closing an app in the right moment, because timer didn't stop working before main method did and timer continue counting. I wish we can test it but I can't imagine different tests than manual with current test suite and with test tools that we use.

vlussenburg:
> > This requires a (unit) test.
> > a) to prove the problem exists (before the code change)
> > b) to verify the problem is fixed
> > and of course c) to show all scenarios we support.
> 
> I wish Unity can provide an easy tool to mock moving device to background and watch background thread time to synchronize them somehow - you were able to reproduce it by just closing an app in the right moment, because timer didn't stop working before main method did and timer continue counting. I wish we can test it but I can't imagine different tests than manual with current test suite and with test tools that we use.

Check. Makes sense.

vlussenburg (approved):

PR #67, opened 2021-03-09
Background thread support

konraddysput:
# Why
Our users would like to start using backtraceClient instance from background threads/coroutines/async methods. In examples that we saw in the past they're sharing an instance of backtraceClient between threads - which is kinda not safe, and use it to send reports when an exception occurs. 

Our users would like to use Backtrace to detect and capture exceptions that background threads generate. 

With this approach, when they will use backtraceClient to send a report from any background thread, because of Unity engine dependency we won't be able to use Unity API to generate the report - for example, we won't be able to start coroutine or generate attributes.

This diff adds background exception cache with the respect of exception information (so everything related to the background thread will be still available). Once Update method fire in the Backtrace client we will be able to send it from main thread.

# Testing strategy

**Send report/exception/message from Backtrace client from background thread**
- previously it generated an exception
- now it stores a report that will be ready to send on the next update method
- Please see unit tests to validate the behavior

**Capture unhandled thread exception generated by background thread in Backtrace**
- previously Backtrace client didn't support it
- now it allows to capture them and use unhandled exception flow like skip report etc.
- Please see unit tests to validate the behavior.

# Example code that previously didn't work

```
    private async Task<WWW> MakeHttpRequest()
    {
        var www = await new WWW("https://jsonplaceholder.typicode.com/usersa");
        return www;

    }

    private async Task DownloadUsers()
    {
        var www = await MakeHttpRequest();


        if (!string.IsNullOrEmpty(www.error))
        {
            throw new Exception(www.error);
        }
    }
    private async Task PushAsyncStateMachineToLimits()
    {
            try
            {
                await DownloadUsers();
            }
            catch (Exception e)
            {
                _client.Send(e);
            }
            return;
    }

    private IEnumerator TestAsyncStateMachine()
    {
        PushAsyncStateMachineToLimits();
        yield return null;
    }
```

or 
```
        var thread = new Thread(() =>
            {
                var exception = new InvalidOperationException(exceptionMessage);
                client.Send(exception);
            });
```


Runtime/BacktraceClient.cs
```diff
@@ -20,14 +21,16 @@ public class BacktraceClient : MonoBehaviour, IBacktraceClient
...
-        public const string VERSION = "3.3.3";
+        public const string VERSION = "3.3.4";
         public bool Enabled { get; private set; }
 
         /// <summary>
         /// Client attributes
         /// </summary>
         private readonly Dictionary<string, string> _clientAttributes = new Dictionary<string, string>();
 
+        internal readonly Stack<BacktraceReport> BackgroundExceptions = new Stack<BacktraceReport>();
```
rqbacktrace, line 32: Why is this Stack instead of Queue?

Runtime/BacktraceClient.cs
```diff
@@ -824,10 +854,22 @@ private bool ShouldSendReport(Exception exception, List<string> attachmentPaths,
...
             bool shouldProcess = _reportLimitWatcher.WatchReport(new DateTime().Timestamp());
             if (shouldProcess)
             {
+                // This condition checks if we should send exception from current thread
+                // if comparision result confirm that we're trying to send an exception from different
+                // thread than main, we should add the exception object to the exception list 
+                // and let update method send data to Backtrace.
+                if (Thread.CurrentThread.ManagedThreadId != _current.ManagedThreadId)
+                {
+                    var report = new BacktraceReport(exception, attributes, attachmentPaths);
```
rqbacktrace, line 868: This seems to be repeated 3 times. Can we pull this into a function?

Runtime/BacktraceClient.cs
```diff
@@ -413,15 +416,28 @@ private void Awake()
         /// <summary>
         /// Update native client internal ANR timer.
         /// </summary>
-        private void Update()
+        private void LateUpdate()
```
rqbacktrace, line 419: Why is this `LateUpdate` now? What happens in `Update`?

rqbacktrace (approved):
Looks good to me! Just had a few questions

Runtime/BacktraceClient.cs
konraddysput: Not really because of different parameters that this function require and cost of BacktraceReport creation

Runtime/BacktraceClient.cs
konraddysput: To give main thread to update before an anr thread

Runtime/BacktraceClient.cs
konraddysput: Hmm yes, you're right. Queue makes more sense here

Runtime/BacktraceClient.cs
rqbacktrace: Ah okay that makes sense

Runtime/BacktraceClient.cs
rqbacktrace: Okay I see. Since this gets called in Update method we need to consider cost of BacktraceReport creation.

vlussenburg (approved):
didn't fully understand the issue but seeing the test made it clear! 👌

PR #68, opened 2021-03-11
Attachment support

konraddysput:
# Why

As a user I would like to upload attachments when:
* ANR,
* Crash,
* Exception,
* Unhandled exception,
happened. With attachment support, we can have more insights about a game state when the specific situation in my game occurred. 

This pull request allows our users to set up attachment support for managed and native reports. To achieve that users can initialize Backtrace Client integration with the list of paths to attachments that they would like to send to us. Attachments will be included in every Backtrace report. In addition to that on iOS when OOM, Crash occurred, we're generating crash/oom attributes but also we store attachments. In the next session when we want to send attachments to Backtrace, old attachments and new attachments will be available in the OOM/Crash report.

**Limitations**
* We can initialize native attachments only on game startup - unfortunately, we can't add more attachments later.
* User is responsible for attachment size - application might crash because of attachment size - for example when user wants to upload 5GB of data to Backtrace as an attachment.

# Updates:
This diff updates the backtrace-android native library available in unity and the backtrace-cocoa (objective-c) library. 

# Testing strategy
In all my tests I used:
* single file (.txt)
* multiple files (.txt and .jpg)
* files that don't exist with files that exist on the hard drive

### iOS 

**Attachments are available on app crash**
1. Set paths to attachments
2. Crash app
3. Relaunch Game 
4. Check the Backtrace report in Backtrace and validate if attachments from the previous session and current session are available.

**OOM reports contain attachment paths**
1. Set paths to attachments
2. Simulate memory warning
3. Crash app via OOM script OR by using special workarounds.
4. Check if the OOM report has attachments from the previous session and from the current session. 

**ANR reports contain attachment paths**
1. Set paths to  attachments
2.  Generate ANR on the managed layer,
3. Check if the ANR report has attachments set by the user in the current session

### Android
**Generate android Crash**
1. Set paths to attachments
2. Crash app 
4. Check the Backtrace report in Backtrace and validate if attachments are available.

**ANR reports contain attachment paths**
1. Set paths to attachments
2.  Generate ANR on the managed layer,
3. Check if the ANR report has attachments set by the user in the current session

## Managed tests
**Send exception/unhandled exception/message and see if attachments are available**
1. Set the path to the attachment
2. Generate an exception/unhandled exception/message
3. Check the Backtrace report in Backtrace to see if


Runtime/BacktraceClient.cs
```diff
@@ -289,6 +292,10 @@ public static BacktraceClient Initialize(BacktraceConfiguration configuration, D
             var backtrackGameObject = new GameObject(gameObjectName, typeof(BacktraceClient), typeof(BacktraceDatabase));
             BacktraceClient backtraceClient = backtrackGameObject.GetComponent<BacktraceClient>();
             backtraceClient.Configuration = configuration;
+            if (attachments != null && attachments.Count != 0)
+            {
+                backtraceClient.Attachments = attachments;
```
rqbacktrace, line 297: Is it safer to add an empty "attachments" object if we don't have any attachments from user?

Runtime/BacktraceClient.cs
```diff
@@ -653,6 +696,7 @@ private BacktraceData SetupBacktraceData(BacktraceReport report)
                 : _backtraceLogManager.ToSourceCode();
 
             report.AssignSourceCodeToReport(sourceCode);
+            report.AttachmentPaths.AddRange(Attachments);
```
rqbacktrace, line 699: Where do we handle managed reports from C# layer?

rqbacktrace: Ah I see, it looks good to me! Thank you!

rqbacktrace (approved):
Looks good to me, thank you!

Runtime/BacktraceClient.cs
konraddysput: We did that - we auto initialize attachments on object startup just after constructor

rqbacktrace: Ah okay I see, thank you!

vlussenburg (commented):
The `.so` files look empty (0 bytes) in the diff - is that right?

Runtime/BacktraceClient.cs
```diff
@@ -289,6 +292,10 @@ public static BacktraceClient Initialize(BacktraceConfiguration configuration, D
             var backtrackGameObject = new GameObject(gameObjectName, typeof(BacktraceClient), typeof(BacktraceDatabase));
             BacktraceClient backtraceClient = backtrackGameObject.GetComponent<BacktraceClient>();
             backtraceClient.Configuration = configuration;
+            if (attachments != null && attachments.Count != 0)
```
vlussenburg, line 295: Realize that when getting passed a `List<String>`, the API user can add paths to it later, since it's an object reference. I'd set the reference regardless, even if it's empty.

If the intent of the API is to pass them and never change them (which is fair) you should clone the list (otherwise you're exposing your privates)

Runtime/Native/iOS/NativeClient.cs
```diff
@@ -77,7 +80,7 @@ public NativeClient(string gameObjectName, BacktraceConfiguration configuration)
         /// Start crashpad process to handle native Android crashes
         /// </summary>
 
-        private void HandleNativeCrashes(BacktraceConfiguration configuration)
+        private void HandleNativeCrashes(BacktraceConfiguration configuration, List<string> attachments)
```
vlussenburg, line 83: Why aren't the attachments part of the configuration object?

Runtime/Native/iOS/NativeClient.cs
```diff
@@ -55,20 +55,23 @@ internal struct Entry
 
 #endif
 
-        public NativeClient(string gameObjectName, BacktraceConfiguration configuration)
+        public NativeClient(BacktraceConfiguration configuration, List<string> attachments)
```
vlussenburg, line 58: Consider adding attachments to configuration object

Runtime/BacktraceClient.cs
```diff
@@ -310,13 +328,26 @@ public static BacktraceClient Initialize(BacktraceConfiguration configuration, D
...
+
+        /// <summary>
+        /// Initialize new Backtrace integration with database path. Note - database path will be auto created by Backtrace Unity plugin
+        /// </summary>
+        /// <param name="url">Server url</param>
+        /// <param name="databasePath">Database path</param>
+        /// <param name="attributes">Client side attributes</param>
+        /// <param name="gameObjectName">game object name</param>
+        /// <returns>Backtrace client</returns>
+        public static BacktraceClient Initialize(string url, string databasePath, Dictionary<string, string> attributes = null, List<string> attachments = null, string gameObjectName = "BacktraceClient")
```
vlussenburg, line 343: Attachments not documented in doc

Runtime/BacktraceClient.cs
```diff
@@ -28,6 +28,8 @@ public class BacktraceClient : MonoBehaviour, IBacktraceClient
         /// </summary>
         private readonly Dictionary<string, string> _clientAttributes = new Dictionary<string, string>();
 
+        public List<string> Attachments { get; internal set; } = new List<string>();
```
vlussenburg, line 31: Generic across the file: I don't like the name `Attachments`. It's not what's in the collection. It's paths to attachments file to upload.

vlussenburg (changes requested):
Although 'boring getter setter logic' this should be unit tested.

Runtime/BacktraceClient.cs
konraddysput: It's unsupported by native libraries that we're using. To prevent this we're making this list readonly and user won't change it over the time or is aware if he will change the reference to the object just after running some kind of api he can't exect this will change also in our class. 

 We allow user to set attachments on startup - I added a note about it on the diff description

Runtime/Native/iOS/NativeClient.cs
konraddysput: How can user define path to attachments in the Untiy IDE - because BacktraceConfiguration class contains a definition of everythign that user see and can edit in the UI. We can right now initialize it only via Initialize method so we can't also add attachments to backtraceConfiguration object because otherwise we user will somehow see them in the UI. For the same reason we're not able to define attributes in the UI.

Runtime/Native/iOS/NativeClient.cs
konraddysput: answer above.

Runtime/Native/iOS/NativeClient.cs
vlussenburg: Hey! I didn't know the Config object was tightly coupled to the UI options - why does that need to be 1:1? Changing something in the UI will have a ripple effect (let's say we want to hide something in the UI and move it to code-only) and that seems like having a HTML field submit name be the same as the database field.

I just base this on the generic "prefer objects to parameters" especially if you already have an object you're passing (for example [here](https://www.codeguru.com/columns/dotnet/using-objects-as-parameters-in-c.htm)). Which is obviously subjective.

Meta: my suggestions are mainly as a passionate ex-developer and you can ignore them if you think they're hogwash :) Up to you, just trying to have add value with my grandpa developer skills :)

vlussenburg (approved):

jasoncdavis0 (approved):

Runtime/Model/BacktraceConfiguration.cs
```diff
@@ -216,15 +217,37 @@ public class BacktraceConfiguration : ScriptableObject
         [Tooltip("If the database is unable to send its record, this setting specifies the maximum number of retries before the system gives up")]
         public int RetryLimit = 3;
 
+        [Tooltip("If the database is unable to send its record, this setting specifies the maximum number of retries before the system gives up")]
```
Dautery91, line 220: I think you copy / pasted the tool tip from the above.  Make sure you update it to reflect the real use of the AttachmentPaths field @konraddysput

Editor/BacktraceConfigurationEditor.cs
```diff
@@ -93,6 +92,13 @@ public override void OnInspectorGUI()
                     EditorGUILayout.HelpBox("Please insert value greater or equal -1", MessageType.Error);
                 }
             }
+
+#if UNITY_ANDROID || UNITY_IOS
+            EditorGUILayout.PropertyField(
+                serializedObject.FindProperty("AttachmentPaths"),
+                new GUIContent(BacktraceConfigurationLabels.LABEL_REPORT_ATTACHMENTS));
```
Dautery91, line 99: @konraddysput does this end up exposing the array as we discussed in the editor?  (The default foldout arrow thing?)  If so, this is good for now!  Side note - as a part of the work I am doing, I am going to try to come up with some custom property drawers to make these easier to use

Dautery91 (changes requested):
@konraddysput - looks great overall.  The only feedback is to change that tooltip (see my comment).  Thanks!

Runtime/Model/BacktraceConfiguration.cs
konraddysput: ah good catch. Thanks

Editor/BacktraceConfigurationEditor.cs
konraddysput: Yes it does - I hope we can make it prettier soon!

Dautery91 (approved):
Looks good, thanks for making the change!

PR #69, opened 2021-03-15
Version 3.4.0 - attachment and background thread support, prevent false-positive anrs

konraddysput:
# Update 3.4.0-rc summary 

- native/managed attachment support - user can add path to attachments via BacktraceClient Initialize method or via Unity Editor Backtrace Window. Once user add them on the initialization method/game object - it will be available in every report that BacktraceClient will send to Backtrace.
- background thread support - now BacktraceClient will report exceptions/unhandled exceptions generated by any background thread.
- ANR detection improvements - BacktraceClient won't report anymore ANRs detected when the application goes to the background.
- `thread.main` attribute support - attribute stores an identifier of a main thread. In this situation user can use value available in this attribute to detect main thread.
- Backtrace native integration will set `_mod_faulting_tid` attribute for ANR reports to set default faulting thread.


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Backtrace Unity Release Notes
 
+## Version 3.4.0
```
jasoncdavis0, line 3: 3.4.0-rc1 ?

konraddysput: good-catch I will update a version.

jasoncdavis0 (commented):
suggest we clarify 3.4.0-rc1 as the version.

vlussenburg:
Tested on iOS
- ANR triggers
- OOM triggers
- native crash triggers
- Managed crash triggers

vlussenburg:
Native attachment tested:
image

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Backtrace Unity Release Notes
 
+## Version 3.4.0-rc1
+- native/managed attachment support - user can add path to attachments via BacktraceClient Initialize method or via Unity Editor Backtrace Window. Once user add them on the initialization method/game object - it will be available in every report that BacktraceClient will send to Backtrace.
```
Dautery91, line 4: Not to be too pedantic here, but you should change "Unity Editor Backtrace Window" to "Unity Editor in the BacktraceClient's inspector.  In Unity:  Inspector <> Window.

Runtime/Model/BacktraceConfiguration.cs
```diff
@@ -222,9 +229,28 @@ public class BacktraceConfiguration : ScriptableObject
...
+            var result = new List<string>();
+            if (AttachmentPaths == null || AttachmentPaths.Length == 0)
+            {
+                return result;
+            }
+
+            foreach (var path in AttachmentPaths)
+            {
+                result.Add(ClientPathHelper.GetFulLPath(path));
+            }
```
Dautery91, line 247: @konraddysput  Corner case you should handle (sorry if I missed it elsewhere:  Check for null / empty items in the AttachmentPaths list.  Given the user can create that list + amend it via the inspector, they could accidentally leave one of the items empty.  I dont think this will throw a null reference exception in that case because I am pretty sure the list would just contain an empty string, but might be worth handling (or at least testing).  Let me know what you think

Might make sense to put this in the Initialize method when you assign the value in the config to the client's AttachmentPath property instead.  Ill leave that up to you.

Runtime/Native/Android/NativeClient.cs
```diff
@@ -301,42 +317,50 @@ public void HandleAnr(string gameObjectName, string callbackName)
...
-                                // update error.type attribute in case when crash happen 
-                                SetAttribute("error.type", "Crash");
+
+                                reported = true;
+                                if (AndroidJNI.AttachCurrentThread() == 0)
+                                {
+                                    // set temporary attribute to "Hang"
+                                    AddAttribute(
+                                        AndroidJNI.NewStringUTF("error.type"),
+                                        AndroidJNI.NewStringUTF("Hang"));
```
Dautery91, line 339: [General Note.  Making a change not necessary for this pull request] Reading through this made me think:  Why are your attribute strings defined in line like this?  In your editor code, you have an entire file where you define all of the Labels in one place, and yet these are in line.  Seems like following a similar pattern would be a good change both from the perspective of organization AND maintaining documentation.  If you have all of the attribute definitions in one place, would be easier for us to ensure any docs are up to date + give users a place in the code to see all of the attributes + quickly link to how they are used.  Let's discuss this as an aside

Dautery91 (commented):
See my comments

Runtime/Model/BacktraceConfiguration.cs
konraddysput: It's handled in different places. User attachment never will be null. Reference:
- user attachments: https://github.com/backtrace-labs/backtrace-unity/blob/753f3a044b7b86386c419adcb76980a75a6b7f06/Runtime/Model/BacktraceConfiguration.cs#L239
- report attachments: https://github.com/backtrace-labs/backtrace-unity/blob/753f3a044b7b86386c419adcb76980a75a6b7f06/Runtime/Model/BacktraceReport.cs#L105

If the list is empty in both examples - it's fine. If list has empty values it's also fine - we will check if file exists before uploading it to Backtrace. We wont want to check if file exists here in this function because user might want to create them after backtrace client initialization.

Runtime/Native/Android/NativeClient.cs
konraddysput: It should be defined in const variables in class definition I agree - but this change might be important/usefull once we will merge implementations for native clients. I think it't an item on my to do list while we will work on native client refactor.

CHANGELOG.md
konraddysput: thanks, Will make a change.

Runtime/Model/BacktraceConfiguration.cs
Dautery91: Ok great, makes sense.  Thanks

konraddysput: @Dautery91 it won't cost us anything to anyway prevents from setting empty strings in editor. I will add anyway a check on the startup. Thanks for pointing this out.

Runtime/Native/Android/NativeClient.cs
Dautery91: Ok sounds good.  Yeah I view this change as primarily a documentation improvement than anything else, so we should just be keeping this in the back of our minds + slot in the work when and where it makes sense.
@vlussenburg

vlussenburg (approved):

PR #70, opened 2021-03-25
Arabic language support

konraddysput:
# Why

Previously we detected an issue with the DateTime format with Arabic langauge.  Application throws an unhandled exception in log manager when backtrace-unity tries to generate a breadcrumb. This diff allows to safe convert DateTime to string with invariant culture support.

Also this diffs adds checks to native iOS attributes - to prevent a situation we will try to convert NULL to string.


rqbacktrace (approved):
Looks good to me!

Dautery91 (approved):

Dautery91:
looks fine on my end

jasoncdavis0 (approved):

PR #72, opened 2021-03-27
Changes to make adhoc custom breadcrumbs possible

vlussenburg:
For some reason (line ending?) the commit looks huge but that was changed is:

`attachment__` to `attachment_` (single underscore)

Backtrace doesn't recognize `_bt-breacrumbs-0`.. needed to get rid of that underscore.


konraddysput:
Please continue here: https://github.com/backtrace-labs/backtrace-unity/pull/73

PR #73, opened 2021-03-29
Correct attachment name

konraddysput:
# Why

attachment name started with underscore prefix `_` . This diff removes unnecessary underscore from Backtrace API implementation.


vlussenburg:
wondering how we would prevent it from accidentally changing again in the future - ideas?

konraddysput:
> wondering how we would prevent it from accidentally changing again in the future - ideas?

We moved it to specific function where we defined constraint for attachment name. Not sure what we can do more if we already defined constraint.

vlussenburg (approved):

PR #74, opened 2021-04-21
Event aggregation logic for reviewers

konraddysput:
This diff shows business logic responsible for event aggregation support in the Backtrace-Unity client. 

Information for reviewers:
* BacktraceClient - class responsible for Backtrace integration in Unity - creates a single instance of BacktraceSession on application startup. Developers can't create more instances than one. BacktraceClient manage BacktraceSession class lifecycle.
*  BacktraceSession allows adding unique (hearthbeat) events and session events.
* BacktraceSession tick method will be invoked by BacktraceClient every time when Unity Engine calls the update method. In Unity, we prefer to avoid timers and this solution prevents from reserving additional resources for timers, jobs etc.
* Diffs also contains integration tests and unit tests. If you have more idea about specific use cases for our users that we should also capture in this diff please let me know!

Code information for reviewers:
* Please focus on the business logic in the code review -  for example: `Why we remove session events after successful upload?`. 

This diff contains only code - no technical descriptions, release announcement, readme/changelog changes. We will handle those changes in different diff once when we will be ready to ship all updates that we want to add to 3.5.0 to our customers

### DIFF
--- a/Runtime/Model/Session/SessionEvent.cs
+++ b/Runtime/Model/Session/SessionEvent.cs
@@ -1,0 +1,30 @@
+﻿using Backtrace.Unity.Json;
+using System.Collections.Generic;
+
+namespace Backtrace.Unity.Model.Session
+{
+    internal sealed class SessionEvent : EventAggregationBase
+    {
+        internal const string MetircsGroupName = "metric_group";
+        internal readonly IDictionary<string, string> Attributes;
+        internal SessionEvent(string name, long timestamp, IDictionary<string, string> attributes) : base(name, timestamp)
+        {
+            Attributes = attributes;
+        }
+
+        internal BacktraceJObject ToJson(IDictionary<string, string> attributes)
+        {
+            if (Attributes != null)
+            {
+                foreach (var attribute in Attributes)
+                {
+                    attributes[attribute.Key] = attribute.Value;
+                }
+            }
+            var jObject = ToBaseObject(attributes);
+            jObject.Add(MetircsGroupName, Name);
+
+            return jObject;
+        }
+    }
+}
--- a/Runtime/Model/Session/SessionEvent.cs
+++ b/Runtime/Model/Session/SessionEvent.cs
@@ -5,7 +5,7 @@ namespace Backtrace.Unity.Model.Session
 {
     internal sealed class SessionEvent : EventAggregationBase
     {
-        internal const string MetircsGroupName = "metric_group";
+        internal const string MetricGroupName = "metric_group";
         internal readonly IDictionary<string, string> Attributes;
         internal SessionEvent(string name, long timestamp, IDictionary<string, string> attributes) : base(name, timestamp)
         {
@@ -22,7 +22,7 @@ namespace Backtrace.Unity.Model.Session
                 }
             }
             var jObject = ToBaseObject(attributes);
-            jObject.Add(MetircsGroupName, Name);
+            jObject.Add(MetricGroupName, Name);
 
             return jObject;
         }
--- a/Runtime/Model/Session/SessionEvent.cs
+++ b/Runtime/Model/Session/SessionEvent.cs
@@ -1,4 +1,5 @@
-﻿using Backtrace.Unity.Json;
+﻿using Backtrace.Unity.Common;
+using Backtrace.Unity.Json;
 using System.Collections.Generic;
 
 namespace Backtrace.Unity.Model.Session
@@ -7,6 +8,9 @@ namespace Backtrace.Unity.Model.Session
     {
         internal const string MetricGroupName = "metric_group";
         internal readonly IDictionary<string, string> Attributes;
+        internal SessionEvent(string name) : this(name, DateTimeHelper.Timestamp(), new Dictionary<string, string>())
+        { }
+
         internal SessionEvent(string name, long timestamp, IDictionary<string, string> attributes) : base(name, timestamp)
         {
             Attributes = attributes;
--- a/Runtime/Services/BacktraceSession.cs
+++ b/Runtime/Services/BacktraceSession.cs
@@ -20,9 +20,9 @@ namespace Backtrace.Unity.Services
         private const string StartupEventName = "Application Launches";
 
         /// <summary>
-        /// Default number of retries
+        /// Default maximum number of attemps
         /// </summary>
-        public const int DefaultNumberOfRetries = 3;
+        public const int MaxNumberOfAttemps = 3;
 
         /// <summary>
         /// Time between 
@@ -32,7 +32,7 @@ namespace Backtrace.Unity.Services
         /// <summary>
         /// Maximum time between requests
         /// </summary>
-        public const int RetryTimeMax = 5 * 60;
+        public const int MaxTimeBetweenRequests = 5 * 60;
 
         /// <summary>
         /// Submission url
@@ -98,8 +98,7 @@ namespace Backtrace.Unity.Services
         /// </summary>
         private object _object = new object();
 
-        private float _lastUpdateInvoke = 0;
-        private readonly Stack<SessionSubmissionJob> _submissionJobs = new Stack<SessionSubmissionJob>();
+        private readonly List<SessionSubmissionJob> _submissionJobs = new List<SessionSubmissionJob>();
 
         /// <summary>
         /// Create new Backtrace session instance
@@ -126,7 +125,7 @@ namespace Backtrace.Unity.Services
         /// </summary>
         public void SendStartupEvent()
         {
-            Send(new UniqueEvent[0], new SessionEvent[1] { new SessionEvent(StartupEventName) }, DefaultNumberOfRetries);
+            SendPayload(new UniqueEvent[0], new SessionEvent[1] { new SessionEvent(StartupEventName) });
         }
 
         /// <summary>
@@ -135,15 +134,9 @@ namespace Backtrace.Unity.Services
         /// <param name="time">Current game time</param>
         public void Tick(float time)
         {
-            _lastUpdateInvoke = time;
             lock (_object)
             {
-                // check submission job
-                while (_submissionJobs.Count != 0 && _submissionJobs.First().NextInvokeTime < time)
-                {
-                    var submissionJob = _submissionJobs.Pop();
-                    Send(submissionJob.UniqueEvents, submissionJob.SessionEvents, submissionJob.NumberOfRetries);
-                }
+                SendPendingSubmissionJobs(time);
             }
             if (_timeIntervalInMs == 0)
             {
@@ -169,59 +162,9 @@ namespace Backtrace.Unity.Services
         /// </summary>
         public void Send()
         {
-            Send(DefaultNumberOfRetries);
-        }
-
-        /// <summary>
-        /// Force BacktraceSession to Send data to Backtrace with retry settings
-        /// </summary>
-        internal void Send(uint numberOfRetries = DefaultNumberOfRetries)
-        {
-            Send(
-               uniqueEvents: UniqueEvents.ToArray(),
-               sessionEvents: SessionEvents.ToArray(),
-               numberOfRetries: numberOfRetries);
-        }
-        private void Send(UniqueEvent[] uniqueEvents, SessionEvent[] sessionEvents, uint numberOfRetries)
-        {
-            if (numberOfRetries == 0)
-            {
-                Debug.LogWarning("Backtrace Session: Cannot send session data to Backtrace due to submission issue.");
-                ClearEvents();
-                return;
-            }
-            if (uniqueEvents.Length + sessionEvents.Length == 0)
-            {
-                return;
-            }
-            var payload = CreateJsonPayload(uniqueEvents, sessionEvents);
-
-            // cleanup existing copy of events
-            ClearEvents();
-
-            // submit data to Backtrace
-            RequestHandler.Post(SubmissionUrl, payload, (long statusCode, bool httpError, string response) =>
-            {
-                if (statusCode == 200)
-                {
-                    OnRequestCompleted();
-                }
-                else if (statusCode == 503)
-                {
-                    _numberOfDroppedRequests++;
-                    if (numberOfRetries - 1 != 0)
-                    {
-                        // assume that we should try to retry request on 503
-                        _submissionJobs.Push(new SessionSubmissionJob()
-                        {
-                            UniqueEvents = uniqueEvents,
-                            SessionEvents = sessionEvents,
-                            NextInvokeTime = CalculateNextRetryTime(DefaultNumberOfRetries - numberOfRetries),
-                            NumberOfRetries = numberOfRetries - 1
-                        });
-                    }
-                }
-            });
+            SendPayload(
+              uniqueEvents: UniqueEvents.ToArray(),
+              sessionEvents: SessionEvents.ToArray());
         }
 
         /// <summary>
@@ -253,17 +196,6 @@ namespace Backtrace.Unity.Services
             UniqueEvents.AddLast(@event);
             return true;
         }
-
-        /// <summary>
-        /// Determine if Backtrace Session can add next event to store
-        /// </summary>
-        /// <param name="name">event name</param>
-        /// <returns>True if we're able to add. Otherwise false.</returns>
-        private bool ShouldProcessEvent(string name)
-        {
-            return !string.IsNullOrEmpty(name) && (_maximumNumberOfEventsInStore == 0 || (Count() + 1 <= _maximumNumberOfEventsInStore));
-        }
-
 
         /// <summary>
         /// Get number of events available right now in store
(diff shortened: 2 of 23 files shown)


Runtime/Model/Session/SessionEvent.cs
```diff
@@ -0,0 +1,30 @@
+﻿using Backtrace.Unity.Json;
+using System.Collections.Generic;
+
+namespace Backtrace.Unity.Model.Session
+{
+    internal sealed class SessionEvent : EventAggregationBase
+    {
+        internal const string MetircsGroupName = "metric_group";
```
ahicks92, line 8: Should probably be `MetricGroupName` (backend is singular, and this is also misspelled).

Runtime/Services/BacktraceSession.cs
```diff
@@ -0,0 +1,317 @@
...
+                RequestHandler.IgnoreSslValidation = value;
+            }
+        }
+
+        private int _numberOfDroppedEvents = 0;
+
+        /// <summary>
+        /// List of unique events that will be added to next session submission payload
+        /// </summary>
+        internal readonly LinkedList<UniqueEvent> UniqueEvents = new LinkedList<UniqueEvent>();
```
ahicks92, line 46: In practice, it is likely that using whatever C#'s equivalent of a vector or array is will perform slightly better, especially if initial capacity is reserved.

Runtime/Services/BacktraceSession.cs
```diff
@@ -0,0 +1,317 @@
...
+        /// </summary>
+        /// <param name="eventName">session event name</param>
+        public bool AddSessionEvent(string eventName, IDictionary<string, string> attributes = null)
+        {
+            if (!ShouldProcessEvent(eventName))
+            {
+                Debug.LogWarning("Skipping report: Reached store limit or event has empty name.");
+                return false;
+            }
+            var existingEvent = SessionEvents.FirstOrDefault(m => m.Name == eventName);
```
ahicks92, line 243: We need to allow sending duplicate events since session events is a counter.

Runtime/Services/BacktraceSession.cs
```diff
@@ -0,0 +1,317 @@
...
+                    Send(numberOfRetries - 1);
+                }
+            });
+        }
+
+        /// <summary>
+        /// Add unique event to next Backtrace session request
+        /// </summary>
+        /// <param name="attributeName">attribute name</param>
+        public bool AddUniqueEventAttribute(string attributeName, IDictionary<string, string> attributes = null)
```
ahicks92, line 180: I'm not 100% sure I know enough about unique events to say for sure, but maybe this should be renamed, since the purpose is to enqueue an event to be sent.  Maybe just `AddUniqueEvent`.

Runtime/Services/BacktraceSession.cs
```diff
@@ -0,0 +1,317 @@
...
+            _attributeProvider.AddAttributes(attributes);
+
+            // validate if unique event attribute is available and
+            // prevent undefined attributes
+            if (attributes.TryGetValue(attributeName, out string attributeValue) == false || string.IsNullOrEmpty(attributeValue))
+            {
+                Debug.LogWarning("Attribute name is not available in attribute scope. Please define attribute to set unique event.");
+                return false;
+            }
+            var existingEvent = UniqueEvents.FirstOrDefault(m => m.Name == attributeName);
```
ahicks92, line 201: The actual logic for this needs to check that all attributes are equal for both events, not just that names match.  For unique, sending two copy of the exactly duplicate event doesn't make sense, but all other cases should likely be allowed (e.g. what if we're using a linked attribute level, and it's changing all the time).

I want to suggest not trying to guard against this at all is the right choice both for unique and for session.

Runtime/Services/BacktraceSession.cs
```diff
@@ -0,0 +1,317 @@
...
+            {
+                return RequestHandler.IgnoreSslValidation;
+            }
+            set
+            {
+                RequestHandler.IgnoreSslValidation = value;
+            }
+        }
+
+        private int _numberOfDroppedEvents = 0;
```
ahicks92, line 41: You probably want `_numberOfDroppedRequests`, since this counts requests, not events.

Runtime/Services/BacktraceSession.cs
```diff
@@ -0,0 +1,317 @@
...
+            CleanEvents();
+
+            // submit data to Backtrace
+            RequestHandler.Post(SubmissionUrl, payload, (long statusCode, bool httpError, string response) =>
+            {
+                if (statusCode == 200)
+                {
+                    OnRequestCompleted();
+                }
+                else
```
ahicks92, line 167: On first read, it seems that this doesn't try with backoff.

Runtime/Services/BacktraceSession.cs
```diff
@@ -0,0 +1,317 @@
...
+                {
+                    OnRequestFailure(uniqueEvents, sessionEvents);
```
ahicks92, line 169: because `OnrRequestFailure` always re-adds these events, doesn't this mean that we effectively retry sending them forever while blocking any further events the user tries to add?

Runtime/Services/BacktraceSession.cs
```diff
@@ -0,0 +1,317 @@
...
+                    // assume that we should try to retry request on 503
```
ahicks92, line 170: Is this comment referring to logic that's going to be in the other prs?  it looks like the current logic retries for any status code that's not exactly 200 (e.g. other 2xx responses also retry, etc).

Runtime/Services/BacktraceSession.cs
```diff
@@ -0,0 +1,317 @@
...
+
+            var @event = new SessionEvent(eventName, DateTimeHelper.Timestamp(), attributes);
+            SessionEvents.AddLast(@event);
+            return true;
+        }
+
+        /// <summary>
+        /// Clean unique and session events stored in Backtrace Session
+        /// </summary>
+        public void CleanEvents()
```
ahicks92, line 257: I think naming this `ClearEvents` would be a better name.

ahicks92 (commented):
I'll give this another pass once we have the application launches event in.  I didn't check the tests.

Runtime/Model/Session/SessionEvent.cs
konraddysput: solved, thanks!

Runtime/Services/BacktraceSession.cs
konraddysput: I'm using LinkedList because I want to add events to the beginning of the queue when http error occurred and we want to retry http request. Normally rest of the events go to the end.

Runtime/Services/BacktraceSession.cs
konraddysput: Your name is much better than mine. Thanks! Renaming method name.

Runtime/Services/BacktraceSession.cs
konraddysput: That's right. Renaming variable. Thanks!

Runtime/Services/BacktraceSession.cs
konraddysput: Renamed. Thanks!

Runtime/Services/BacktraceSession.cs
konraddysput: Based on our Slack thread available here: https://backtrace.slack.com/archives/C01NYBYN7MY/p1618248208041800 we wanted to:
1. avoid retries (Samy)
2. retry on 503

I thought retry on at least 503 give us something. 

I think we will review retry logic later so this part of the pull request might change after first round of feedback from our customers/tech services/ other

Runtime/Services/BacktraceSession.cs
konraddysput: You're right. We can make this logic simpler. I will expose send and retry logic to another function to make it cleaner for other developers to read this code. Thanks!

Runtime/Services/BacktraceSession.cs
konraddysput: Thanks. Removed this check. 

Let me also clarify - if we have two the same session events, maybe we should use some kind of counter instead of putting two the same events in session events array?

Runtime/Services/BacktraceSession.cs
konraddysput: Removed unique check - somehow I though we should send only one unique/session event with specific name. Thanks for pointing this out!

Runtime/Services/BacktraceSession.cs
konraddysput: Since we're not in API that allows us to use coroutines (mono behavior) we can't, for example, wait for n seconds without blocking the main thread. We shouldn't also use threads because we previously agreed that this will increase resources that backtrace-unity integration consumes and might impact games with a small pc budget. unfortunately, I can't have a better solution for exponential backoff than trying instantly just after the previous request :(

Runtime/Services/BacktraceSession.cs
ahicks92: Maybe, but currently the service isn't set up for it and telling when two events are different is ust involved enough that I'd rather not.  I would expect that the number of events from an individual client is going to be low in practice.

Runtime/Services/BacktraceSession.cs
ahicks92: Given the volume in question, we can't afford for these to spin out if the service goes down.  We can wait for @wca to chime in if we want but this will effectively end up DDOSing ourselves one day.  if we have to use a thread, then we probably have to use a thread.  Maybe we can only do it when automatic submission is enabled?

I see that the interface for requests uses a callback; do we have a way to also do callbacks on timers, e..g JS setTimeout or something?

konraddysput: Hey. I've moved this code to tick/update method. Tick/Update method will be invoked by Untiy engine every time when unity generates a new frame. So tick method will occur every time when unity will try to render a new frame. With that being said without an additional thread/timer we can use the update method to schedule jobs to be sent by BacktraceSession.

Now I added default timeout which is 10 seconds. Every time when we failed we're waiting at least 10 seconds for the next retry. Also like previously - we're making only 3 requests. 

@ahicks92 what do you think about this solution?

Tests/Runtime/Session/BacktraceSessionTests.cs
```diff
@@ -0,0 +1,282 @@
...
+            const int expectedNumberOfEvents = 2;
+            var backtraceSession = new BacktraceSession(_attributeProvider, _submissionUrl, 1, maximumNumberOfEvents);
+            var requestHandler = new BacktraceHttpClientMock();
+            backtraceSession.RequestHandler = requestHandler;
+            for (int i = 0; i < expectedNumberOfEvents; i++)
+            {
+                backtraceSession.AddSessionEvent($"{SessionEventName} {i}");
+            }
+
+            backtraceSession.Tick(0);
```
rqbacktrace, line 183: Why tick 0?

Tests/Runtime/Session/BacktraceSessionTests.cs
```diff
@@ -0,0 +1,282 @@
...
+        [Test]
+        public void BacktraceSessionDefaultEvent_ShouldSendDefaultEventOnTheApplicationStartup_DataWasSendToBacktrace()
+        {
+            var backtraceSession = new BacktraceSession(_attributeProvider, _submissionUrl, 10, 3);
+            var requestHandler = new BacktraceHttpClientMock();
+            backtraceSession.RequestHandler = requestHandler;
+
+            backtraceSession.SendStartupEvent();
+
+            Assert.AreEqual(1, requestHandler.NumberOfRequests);
```
rqbacktrace, line 278: Should we validate any `StartupEvent` params?

Tests/Runtime/Session/BacktraceSessionTests.cs
```diff
@@ -0,0 +1,282 @@
...
+
+        [Test]
+        public void BacktraceSession_ShouldTriggerUploadProcessOnSendMethodWithOnlyUniqueEvent_DataWasSendToTheService()
+        {
+            var jsonString = string.Empty;
+            var submissionUrl = string.Empty;
+            var backtraceSession = new BacktraceSession(_attributeProvider, _submissionUrl, 0, 1);
+            backtraceSession.RequestHandler = new BacktraceHttpClientMock()
+            {
+                OnIvoke = (string url, BacktraceJObject json) =>
```
rqbacktrace, line 52: Typo: `OnIvoke` -> `OnInvoke`

Runtime/Services/BacktraceSession.cs
```diff
@@ -0,0 +1,354 @@
...
+        /// <summary>
+        /// Last update time that will be updated after each update start
+        /// </summary>
+        private float _lastUpdateTime = 0;
+
+        /// <summary>
+        /// Maximum number of events in store. If number of events in store hit the limit
+        /// BacktraceSession instance will send data to Backtrace.
+        /// </summary>
+        private uint _maximumNumberOfEventsInStore;
```
rqbacktrace, line 83: Should Unique and Session events share the same max?

Runtime/Services/BacktraceSession.cs
```diff
@@ -0,0 +1,354 @@
...
+            // add unique events
+            var uniqueEventsJson = new List<BacktraceJObject>();
+            foreach (var uniqueEvent in uniqueEvents)
+            {
+                uniqueEventsJson.Add(uniqueEvent.ToJson());
+            }
+
+            jsonData.Add("unique_events", uniqueEventsJson);
+
+            // add unique events
```
rqbacktrace, line 325: Typo in comment, should be `add session events`

rqbacktrace (approved):
Looks good to me! Just had some minor comments!

Tests/Runtime/Session/BacktraceSessionTests.cs
```diff
@@ -0,0 +1,282 @@
...
+                    submissionUrl = url;
+                }
+            };
+
+            backtraceSession.AddSessionEvent(SessionEventName);
+            backtraceSession.Send();
+
+            Assert.IsNotEmpty(jsonString);
+            Assert.AreEqual(submissionUrl, _submissionUrl);
+            Assert.IsEmpty(backtraceSession.SessionEvents);
```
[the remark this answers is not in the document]
konraddysput: We have other separated tests responsible for teting json serialization. Unfortunately we don't have custom json deserialization and json deserialization available in Untiy doesn't support attributes.

Tests/Runtime/Session/BacktraceSessionTests.cs
```diff
@@ -0,0 +1,282 @@
...
+
+        [Test]
+        public void BacktraceSession_ShouldTriggerUploadProcessOnSendMethodWithOnlySessionEvent_DataWasSendToTheService()
+        {
+            var jsonString = string.Empty;
+            var submissionUrl = string.Empty;
+            var backtraceSession = new BacktraceSession(_attributeProvider, _submissionUrl, 0, 1);
+            backtraceSession.RequestHandler = new BacktraceHttpClientMock()
+            {
+                OnIvoke = (string url, BacktraceJObject json) =>
```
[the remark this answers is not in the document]
konraddysput: done! Thanks

Tests/Runtime/Session/BacktraceSessionTests.cs
```diff
@@ -0,0 +1,282 @@
...
+                var time = BacktraceSession.RetryTimeMax + (BacktraceSession.RetryTimeMax * i) + i + 1;
+                backtraceSession.Tick(time);
+            }
+
+            yield return new WaitForSeconds(1);
+            Assert.AreEqual(BacktraceSession.DefaultNumberOfRetries, requestHandler.NumberOfRequests);
+        }
+
+        [Test]
+        public void BacktraceSession_ShouldTryAtLeastOneOnHttpFailure_DataWasntSendToBacktrace()
```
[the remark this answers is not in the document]
konraddysput: Renamed, thanks!

Tests/Runtime/Session/BacktraceSessionTests.cs
konraddysput: 0 means current time - We have value 0 on the application startup. For testing purpose it was enough because we wanted to force upload process when we have enough data to send or (in this example) test if we will send anything when we shouldn't.

Tests/Runtime/Session/BacktraceSessionTests.cs
konraddysput: It's constant and unfortunately this function won't store anything on the event queue - it will create in memory a json object and send it : (

Tests/Runtime/Session/BacktraceSessionTests.cs
konraddysput: done! Thanks!

Runtime/Services/BacktraceSession.cs
konraddysput: _maximumNumberOfEventsInStore = number of events available in unique events store and session events store. For reference let me copy paste here a ShouldProcessEvent function:
```
 private bool ShouldProcessEvent(string name)
        {
            return !string.IsNullOrEmpty(name) && (_maximumNumberOfEventsInStore == 0 || (Count() + 1 <= _maximumNumberOfEventsInStore));
        }
```
and Count implementation
```
     public int Count()
        {
            return UniqueEvents.Count + SessionEvents.Count;
        }

```
as you can see I'm comparing number of events in both queues with this variable.

Runtime/Services/BacktraceSession.cs
konraddysput: Thanks!

Runtime/Services/BacktraceSession.cs
rqbacktrace: I understand. I just wasn't sure if maybe Unique events and Session events should have their own maxes, like
`_maximumNumberOfUniqueEvents` and `_maximumNumberOfSessionEvents`? Maybe user wants to set different limits for different event types? Maybe we don't care about that use case though which is fine.

Runtime/Services/BacktraceSession.cs
```diff
@@ -0,0 +1,354 @@
...
+            payload.Add("dropped_events", _numberOfDroppedRequests);
+            return payload;
+        }
+
+        private double CalculateNextRetryTime(uint numberOfRetries)
+        {
+            const int jitterFraction = 1;
+            const int backoffBase = 10;
+            var value = DefaultTimeInSecBetweenRequests * Math.Pow(backoffBase, numberOfRetries);
+            var retryLower = Math.Max(0, Math.Min(RetryTimeMax, value));
```
ahicks92, line 349: This is where you'd want clamp.

Runtime/Services/BacktraceSession.cs
```diff
@@ -0,0 +1,354 @@
...
+            var retryUpper = retryLower + retryLower * jitterFraction;
+            return new System.Random().NextDouble() * (retryUpper - retryLower) + retryLower;
```
ahicks92, line 351: Fine if we don't have one, but this would be clearer if there were a version of random that took the endpoints of the range, rather than requiring us to scale.

Runtime/Services/BacktraceSession.cs
```diff
@@ -0,0 +1,354 @@
...
+        }
+
+        private BacktraceJObject CreatePayloadMetadata()
+        {
+            var payload = new BacktraceJObject();
+            payload.Add("dropped_events", _numberOfDroppedRequests);
+            return payload;
+        }
+
+        private double CalculateNextRetryTime(uint numberOfRetries)
```
ahicks92, line 344: I'd put this closer to send.

Runtime/Services/BacktraceSession.cs
```diff
@@ -0,0 +1,354 @@
...
+                }
+                _lastUpdateTime = time;
+            }
+            Send();
+        }
+
+        /// <summary>
+        /// Force BacktraceSession to Send data to Backtrace
+        /// </summary>
+        public void Send()
```
ahicks92, line 170: I'm not a fan of how we're doing recursion by overloading a function of the same name 3 times.  I'd consider giving these unique names.

Runtime/Services/BacktraceSession.cs
konraddysput: I think we don't care about this use case since we send them together and this option is mostly to force sending data when we hit the limit.

Runtime/Services/BacktraceSession.cs
```diff
@@ -0,0 +1,354 @@
...
+            ClearEvents();
+
+            // submit data to Backtrace
+            RequestHandler.Post(SubmissionUrl, payload, (long statusCode, bool httpError, string response) =>
+            {
+                if (statusCode == 200)
+                {
+                    OnRequestCompleted();
+                }
+                else if (statusCode == 503)
```
ahicks92, line 209: There's a pretty good chance this will be "retry on all 5xx" instead.

Runtime/Services/BacktraceSession.cs
```diff
@@ -0,0 +1,354 @@
...
+                {
+                    _numberOfDroppedRequests++;
+                    if (numberOfRetries - 1 != 0)
```
ahicks92, line 212: This code is where you want your retry counter to go up instead of down.  If it did, you can just compare directly to `defaultNumberOfRetries` after incrementing, and pass directly to `CalculateNextRetryTime`.  it seems like it would be easy to forget whether or not the number of retries is going up or going down, and almost anything interesting you want to do with retry counters needs them to be going up so I'd rather see us standardize on that.

ahicks92 (changes requested):
Only really reviewed the retry/backoff logic.

Runtime/Services/BacktraceSession.cs
konraddysput: Unity uniform function require to use floats. Here we have doubles. So I prefer to stay with this implementation instead of just casting it to floats.

Runtime/Services/BacktraceSession.cs
konraddysput: Moved it closer to the Send method.

Runtime/Services/BacktraceSession.cs
konraddysput: How about renaming `Send` method with unique/session events parameter to `SendPayload`?

Runtime/Services/BacktraceSession.cs
konraddysput: renamed to attemps and now we're increasing variable (not decreasing it like previously)

Runtime/Services/BacktraceSession.cs
konraddysput: statusCode > 501 && statusCode != 505

Runtime/Services/BacktraceSession.cs
konraddysput: Moved it to our internal helpers - Unity Clamp doesn't support doubles.

Runtime/Services/BacktraceSession.cs
konraddysput: Moved to Internal helpers (like Clamp)

ahicks92 (approved):

ahicks92:
Approved pending changing the last place where we're flipping attempts which I discussed on slack.

PR #79, opened 2021-04-30
Event aggregation support

konraddysput:
This diff shows event aggregation support. In the previous diff  (https://github.com/backtrace-labs/backtrace-unity/pull/74) we reviewed a business logic in the backtrace-unity plugin. This diff shows all changes that we've added to the backtrace-unity plugin to add support for event aggregation support. Diff contains:
* attribute refactor,
* http client refactor,
* event aggregation support

Information for reviewers:
* In this diff we should extend our readme and changelog with information about event aggregation support. @Dautery91 @vlussenburg  can you help me with it?
* In this diff we should also work on the UI changes that will be available in the Unity Editor
* @rqbacktrace could you please review parts of the diff that you didn't see in the previous review?


Runtime/BacktraceClient.cs
```diff
@@ -26,11 +27,16 @@ public class BacktraceClient : MonoBehaviour, IBacktraceClient
...
+        /// Client attribute provider
         /// </summary>
-        private readonly Dictionary<string, string> _clientAttributes = new Dictionary<string, string>();
+        private AttributeProvider _attributeProvider;
 
-        internal readonly Stack<BacktraceReport> BackgroundExceptions = new Stack<BacktraceReport>();
+        /// <summary>
+        /// Backtrace session instance
+        /// </summary>
+        public IBacktraceSession Session { get; internal set; }
```
vlussenburg, line 37: Need import `using Backtrace.Unity.Interfaces;`

Runtime/Interfaces/IBacktraceSession.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Collections.Generic;
```
vlussenburg, line 1: Needs meta file

Editor/BacktraceConfigurationEditor.cs
```diff
@@ -93,11 +95,37 @@ public override void OnInspectorGUI()
...
+                        eventAggregationSubmissionUrl,
+                        new GUIContent(BacktraceConfigurationLabels.LABEL_EVENT_AGGREGATION_URL));
+
+                    if (!BacktraceConfiguration.ValidateServerUrl(eventAggregationSubmissionUrl.stringValue))
+                    {
+                        EditorGUILayout.HelpBox("Please insert valid Backtrace server url!", MessageType.Error);
+                    }
+
+                    EditorGUILayout.PropertyField(
+                        serializedObject.FindProperty("TimeIntervalInMs"),
```
vlussenburg, line 120: FWIW this line creates an exception rn in my Unity editor

Runtime/Extensions/UnityWebReqeustExtensions.cs
```diff
@@ -0,0 +1,39 @@
+﻿using Backtrace.Unity.Model;
+using System.Text;
+using UnityEngine.Networking;
+
+[assembly: System.Runtime.CompilerServices.InternalsVisibleTo("Backtrace.Unity.Tests.Runtime")]
+namespace Backtrace.Unity.Extensions
+{
+    internal static class UnityWebReqeustExtensions
```
vlussenburg, line 8: typo Reqeust

Runtime/Model/Attributes/IScopeAttributeProvider.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Collections.Generic;
+
+namespace Backtrace.Unity.Model.Attributes
+{
+    internal interface IScopeAttributeProvider
```
vlussenburg, line 5: Great addition / way to group attributes!

Editor/BacktraceConfigurationEditor.cs
konraddysput: It looks like an issue on my side when I merged two different branches. I will take a look and fix it tomorrow. Sorry for difficulties

vlussenburg (commented):
I think I've encountered a bug. I have this game code:
```
    void Awake()
    {
        AsteraX.backtraceClient = GetComponent<BacktraceClient>();
        AsteraX.backtraceClient["backtrace-unity-commit-sha"] = "1e2d885e9cc038b69bea41fb2be59fb1dbc02600";
```

That used to work, but now I get:

```
NullReferenceException: Object reference not set to an instance of an object
Backtrace.Unity.BacktraceClient.set_Item (System.String index, System.String value) (at Library/PackageCache/io.backtrace.unity@1e2d885e9c/Runtime/BacktraceClient.cs:57)
AsteraX.Awake () (at Assets/__Scripts/AsteraX.cs:81)
```

Runtime/Model/Session/SessionSubmissionJob.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Backtrace.Unity.Model.Session
+{
+    internal sealed class SessionSubmissionJob
+    {
+        public double NextInvokeTime { get; set; }
+        public UniqueEvent[] UniqueEvents { get; set; }
+        public SessionEvent[] SessionEvents { get; set; }
+        public uint NumberOfAttemps { get; set; }
```
Dautery91, line 8: @konraddysput spelling mistake

Runtime/Extensions/UnityWebReqeustExtensions.cs
konraddysput: done! Thx

Runtime/Services/BacktraceMetrics.cs
```diff
@@ -0,0 +1,408 @@
...
+    {
+        /// <summary>
+        /// Default time interval in min
+        /// </summary>
+        public const uint DefaultTimeIntervalInMin = 30;
+
+        /// <summary>
+        /// Default time interval in ms
+        /// </summary>
+        public const uint DefaultTimeIntervalInMs = DefaultTimeIntervalInMin * 60;
```
vlussenburg, line 25: That's in seconds not in ms

konraddysput: Adjusted.

Runtime/Interfaces/IBacktraceMetrics.cs
```diff
@@ -0,0 +1,41 @@
+﻿using Backtrace.Unity.Model.Metrics;
+using System.Collections.Generic;
+
+namespace Backtrace.Unity.Interfaces
+{
+    public interface IBacktraceMetrics
+    {
+        /// <summary>
+        /// This list contains the unique events which will be sent whenever Send is triggered (manually or automatically). 
+        /// By default, this list contains the guid attribute. This list is mutable, which means the default can be removed
```
rqbacktrace, line 10: This list is mutable but what does it mean to mutate this list since we can only `get` it?

konraddysput: List has.

rqbacktrace: I agree. But why do we even need to `get` this then? Does the user care about using this list?

konraddysput: In special case if user has to manipulate list - do any operations on them like for example removing them

rqbacktrace: But this list is just a copy right? I think changing this list won't really do anything

konraddysput: Its not a copy. Its a reference. You can manipulate objects three. However we'rr changing unique events and this wont be available

rqbacktrace: Is there some kind of c# language trick here? It looks like we are just using the c# get/set syntax here which returns a value not a reference? I just want to understand, I know it doesn't really matter because we're changing this anyway

rqbacktrace: Okay nevermind, I did some more reading and I understand now. You can ignore this lol thank you!

PR #80, opened 2021-04-30
iOS structure size change

konraddysput:
**Why**

Previously we assumed that structure size in the iOS is constant. Based on the feedback that we got from our customers we should be prepared for a situation when structure size is different. To achieve that we should start using IntPtr.Size - instead of assuming struct size has always the same size = 16.


ianrice07 (approved):
Looks good to me.  Seems to be fixed based on feedback from other users.

PR #81, opened 2021-04-30
Guess native library path when application context is not available

konraddysput:
**Why**

We detected that sometimes when the Unity Android solution is being generated by 3rd party plugin, the application context object might not be available on the application startup. In this situation, the plugin currently will throw an unhandled exception and won't initialize native integration. To prevent this situation we added an additional check for nullable application context and also added a code that still allows us to guess library path based on the most popular application setup.


ianrice07 (approved):
This works from when I tested it locally as well as when another group tested it.

PR #85, opened 2021-05-13
Removed warnings generated by Unity2020 + Unity2021

konraddysput:
Removed warnings generated by Unity2020 + Unity2021

### DIFF
--- a/Editor/Native/Android/SymbolsUpload.cs
+++ b/Editor/Native/Android/SymbolsUpload.cs
@@ -1,5 +1,6 @@
 #if UNITY_2019_2_OR_NEWER && UNITY_ANDROID
 using Backtrace.Unity.Model;
+using Backtrace.Unity.Extensions;
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -190,12 +191,7 @@ namespace Backtrace.Unity.Editor.Build
                     {
                         EditorUtility.DisplayProgressBar("Backtrace symbols upload", "Symbols upload progress:", request.uploadProgress);
                     }
-#if UNITY_2020_2_OR_NEWER
-                    var failure = request.result == UnityWebRequest.Result.ConnectionError || request.result == UnityWebRequest.Result.ProtocolError;
-#else
-                request.isNetworkError || request.isHttpError;
-#endif
-                    if (failure)
+                    if (request.ReceivedNetworkError())
                     {
                         Debug.LogWarning(string.Format("Cannot upload symbols to Backtrace. Reason: {0}", request.downloadHandler.text));
                         return;
--- a/Editor/Native/Android/SymbolsUpload.cs
+++ b/Editor/Native/Android/SymbolsUpload.cs
@@ -25,8 +25,14 @@ namespace Backtrace.Unity.Editor.Build
 
         public void OnPostprocessBuild(BuildReport report)
         {
+            var disabledSymbols =
+#if UNITY_2021_1_OR_NEWER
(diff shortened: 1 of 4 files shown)


Editor/Native/Android/SymbolsUpload.cs
```diff
@@ -184,7 +190,11 @@ private void UploadSymbols(string serverUrl, string symbolsToken, string symbols
                     {
                         EditorUtility.DisplayProgressBar("Backtrace symbols upload", "Symbols upload progress:", request.uploadProgress);
                     }
-                    var failure = request.isNetworkError || request.isHttpError;
+#if UNITY_2020_2_OR_NEWER
+                    var failure = request.result == UnityWebRequest.Result.ConnectionError || request.result == UnityWebRequest.Result.ProtocolError;
```
rqbacktrace, line 194: Do we need to move the `var failure` outside of the #ifdefs like with `var disabledSymbols`? It looks like the #else condition here doesn't assign anything

Runtime/Model/BacktraceHttpClient.cs
```diff
@@ -47,7 +47,11 @@ public void Post(string submissionUrl, BacktraceJObject jObject, Action<long, bo
             asyncOperation.completed += (AsyncOperation operation) =>
             {
                 var statusCode = request.responseCode;
-                var isNetworkError = request.isNetworkError || request.isHttpError;
+#if UNITY_2020_2_OR_NEWER
+                var isNetworkError = request.result == UnityWebRequest.Result.ConnectionError || request.result == UnityWebRequest.Result.ProtocolError;
```
rqbacktrace, line 51: We might need to move `var isNetworkError` outside of this #ifdef, otherwise the #else condition doesn't assign anything

Runtime/Services/BacktraceApi.cs
```diff
@@ -202,6 +208,11 @@ public IEnumerator Send(string json, List<string> attachments, Dictionary<string
             {
                 yield return request.SendWebRequest();
                 BacktraceResult result;
+#if UNITY_2020_2_OR_NEWER
+                var networkError = request.result == UnityWebRequest.Result.ConnectionError || request.result == UnityWebRequest.Result.ProtocolError;
```
rqbacktrace, line 212: We might need to move `var networkError` outside of this #ifdef, otherwise the #else condition doesn't assign anything. Also consider renaming to `isNetworkError` for consistency

rqbacktrace (changes requested):
I think it might not compile if `UNITY_2020_2_OR_NEWER` is false

Runtime/Extensions/UnityWebRequestExtensions.cs
```diff
@@ -15,6 +15,16 @@ internal static UnityWebRequest SetMultipartFormData(this UnityWebRequest source
             return source;
         }
 
+        public static bool ReceivedNetworkError(this UnityWebRequest request)
+        {
+            return
+#if UNITY_2020_2_OR_NEWER
+                 request.result == UnityWebRequest.Result.ConnectionError || request.result == UnityWebRequest.Result.ProtocolError;
```
rqbacktrace, line 22: Formatting not aligned

rqbacktrace (approved):
Awesome, this new solution looks much cleaner!

Runtime/Extensions/UnityWebRequestExtensions.cs
konraddysput: Formatted thanks.

PR #87, opened 2021-05-18
Handle different unity player java class name

konraddysput:
**Why**

In some builds, Unity player might be available under a different name or might be overall unavailable for backtrace-unity. This change allows us to use a custom function to guess the native library path and prevents backtrace-unity from throwing an exception on native-client initialization.


rqbacktrace (approved):
Looks good to me! I'm curious what will happen if `GetNativeDirectoryPath()` returns an empty string.

konraddysput:
> Looks good to me! I'm curious what will happen if `GetNativeDirectoryPath()` returns an empty string.

We're guessing lib path based on the pattern that we see in unity

PR #88, opened 2021-05-19
Metrics support imrovements

konraddysput:
This diff allows us to send data to two separated endpoints ( unique + summed events)

### DIFF
--- a/Runtime/Model/BacktraceConfiguration.cs
+++ b/Runtime/Model/BacktraceConfiguration.cs
@@ -266,7 +266,35 @@ namespace Backtrace.Unity.Model
             return result;
         }
 
-        public string GetEventAggregationUrl()
+        public string GetUniverseName()
+        {
+            var submissionUrl = GetValidServerUrl();
+            var submitUrl = submissionUrl.Contains("submit.backtrace.io");
+            if (submitUrl)
+            {
+                const int tokenLength = 64;
+                // we want to skip the last `/`. Since we're counting from 0 we need to decrease 
+                // position by 2
+                var endPosition = submissionUrl.LastIndexOf("/") - tokenLength - 2;
+                var startPosition = submissionUrl.LastIndexOf('/', endPosition) + 1;
+                return submissionUrl.Substring(startPosition, endPosition - startPosition + 1);
+            }
+            else
+            {
+                const string backtraceDomain = "backtrace.io";
+                var domainIndex = submissionUrl.IndexOf(backtraceDomain);
+                if (domainIndex == -1)
+                {
+                    throw new ArgumentException("Invalid Backtrace url");
+                }
+
+                var uri = new UriBuilder(submissionUrl);
+                return uri.Host.Substring(0, uri.Host.IndexOf("."));
+            }
+
+        }
+
+        public string GetToken()
         {
             const int tokenLength = 64;
             const string tokenQueryParam = "token=";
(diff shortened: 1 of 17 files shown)


Runtime/Model/Metrics/UniqueEventsSubmissionQueue.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        internal override IEnumerable<BacktraceJObject> GetEventsPayload(ICollection<UniqueEvent> events)
+        {
+            var uniqueEventsJson = new List<BacktraceJObject>();
+            foreach (var uniqueEvent in events)
+            {
+                uniqueEventsJson.Add(uniqueEvent.ToJson());
+                uniqueEvent.UpdateTimestamp(DateTimeHelper.Timestamp(), _attributeProvider.GenerateAttributes());
+            }
+
```
rqbacktrace, line 41: Why don't we need to clear the events here like we do with SummedEvents? I'm sure it has something to do with the complicated spec lol, it will just help my understanding when I do these implementations myself

Runtime/Services/BacktraceMetrics.cs
```diff
@@ -38,30 +42,47 @@ internal sealed class BacktraceMetrics : IBacktraceMetrics
...
+        /// Summed events submission queue
         /// </summary>
-        public const int MaxTimeBetweenRequests = 5 * 60;
+        internal readonly SummedEventSubmissionQueue _summedEventSubmissionQueue;
+
+        /// <summary>
+        /// Startup event name that will be send on the application startup
+        /// </summary>
+        private const string StartupEventName = "session";
+
```
rqbacktrace, line 70: Extra newline

rqbacktrace (approved):
Looks good to me! Just a few typos but don't want to block on that

Runtime/Model/BacktraceConfiguration.cs
```diff
@@ -266,15 +266,43 @@ public List<string> GetAttachmentPaths()
             return result;
         }
 
-        public string GetEventAggregationUrl()
+        public string GetUniverseName()
+        {
+            var submissionUrl = GetValidServerUrl();
```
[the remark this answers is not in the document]
konraddysput: It will be a new chagne for us and will require new information.

Runtime/Model/Metrics/EventAggregationBase.cs
```diff
@@ -14,7 +14,6 @@ public EventAggregationBase(string name, long timestamp)
             Name = name;
             Timestamp = timestamp;
         }
-
```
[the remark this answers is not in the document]
konraddysput: Added

Runtime/Model/Metrics/MetricsSubmissionJob.cs
```diff
@@ -1,10 +1,11 @@
...
+namespace Backtrace.Unity.Model.Metrics
 {
-    internal sealed class MetricsSubmissionJob
+    internal sealed class MetricsSubmissionJob<T>
     {
         public double NextInvokeTime { get; set; }
-        public UniqueEvent[] UniqueEvents { get; set; }
-        public SummedEvent[] SummedEvents { get; set; }
+        public ICollection<T> Events { get; set; }
         public uint NumberOfAttemps { get; set; }
```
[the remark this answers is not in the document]
konraddysput: done, thx!

Runtime/Model/Metrics/UniqueEventsSubmissionQueue.cs
konraddysput: we don't send the same summed events every time (like heartbeat/unique/(you know the right name ( ͡° ͜ʖ ͡°))

Runtime/Services/BacktraceMetrics.cs
konraddysput: done, thx!

PR #89, opened 2021-05-24
Metrics - allow overriding submission URL

konraddysput:
**Why**

As a user in the on-premise environment, I would like to override the default submission URL that BacktraceMetrics is using. 

This diff allows to override default submission URL and cleanup names in code. Changes:
* removed universe name and token from the BacktraceMetrics constructor,
* allow for overriding default submission URL via setters
* Request handler/urls cleanup
* renamed variables in BacktraceConfiguration - now we using "metrics' name instead of "event aggregation"


rqbacktrace (approved):
Looks good to me!

PR #91, opened 2021-06-07
Screenshot management

konraddysput:
## Why

As a user, I would like to define attachments settings that allow the game to save resources/time when the game generates image attachment.

As a power user, I would like to use database API to define screenshot settings.

### DIFF
--- a/Runtime/BacktraceDatabase.cs
+++ b/Runtime/BacktraceDatabase.cs
@@ -69,7 +69,7 @@ namespace Backtrace.Unity
 
         /// <summary>
         /// Attachment support: Screenshot max height - based on screenshot max height, algorithm calculates
-        /// ration, that allows to calculate screenshot max width
+        /// ratio, that allows to calculate screenshot max width
         /// </summary>
         public int ScreenshotMaxHeight
         {
--- a/Runtime/BacktraceDatabase.cs
+++ b/Runtime/BacktraceDatabase.cs
@@ -52,6 +52,37 @@ namespace Backtrace.Unity
         /// </summary>
         public string DatabasePath { get; protected set; }
 
+        /// <summary>
+        /// Attachment support: Screenshot quality
+        /// </summary>
+        public int ScreenshotQuality
+        {
+            get
+            {
+                return BacktraceDatabaseFileContext.ScreenshotQuality;
+            }
+            set
+            {
+                BacktraceDatabaseFileContext.ScreenshotQuality = value;
+            }
+        }
+
+        /// <summary>
+        /// Attachment support: Screenshot max height - based on screenshot max height, algorithm calculates
+        /// ration, that allows to calculate screenshot max width
+        /// </summary>
+        public int ScreenshotMaxHeight
+        {
+            get
+            {
+                return BacktraceDatabaseFileContext.ScreenshotMaxHeight;
+            }
+            set
+            {
+                BacktraceDatabaseFileContext.ScreenshotMaxHeight = value;
+            }
+        }
+
 
         /// <summary>
         /// Backtrace database instance.
--- a/Runtime/Interfaces/IBacktraceDatabase.cs
+++ b/Runtime/Interfaces/IBacktraceDatabase.cs
@@ -15,6 +15,17 @@ namespace Backtrace.Unity.Interfaces
     public interface IBacktraceDatabase
     {
         /// <summary>
+        /// Screenshot quality
+        /// </summary>
+        int ScreenshotQuality { get; set; }
+
+        /// <summary>
+        /// Screenshot max height - based on screenshot max height, algorithm calculates
+        /// ration, that allows to calculate screenshot max width
+        /// </summary>
+        int ScreenshotMaxHeight { get; set; }
+
+        /// <summary>
         /// Backtrace Breadcrumbs
         /// </summary>
         IBacktraceBreadcrumbs Breadcrumbs { get; }
--- a/Runtime/Interfaces/IBacktraceDatabase.cs
+++ b/Runtime/Interfaces/IBacktraceDatabase.cs
@@ -21,7 +21,7 @@ namespace Backtrace.Unity.Interfaces
 
         /// <summary>
         /// Screenshot max height - based on screenshot max height, algorithm calculates
-        /// ration, that allows to calculate screenshot max width
+        /// ratio, that allows to calculate screenshot max width
         /// </summary>
         int ScreenshotMaxHeight { get; set; }
 
--- a/Runtime/Interfaces/IBacktraceDatabaseFileContext.cs
+++ b/Runtime/Interfaces/IBacktraceDatabaseFileContext.cs
@@ -8,6 +8,17 @@ namespace Backtrace.Unity.Interfaces
     internal interface IBacktraceDatabaseFileContext
     {
         /// <summary>
+        /// Screenshot quality
+        /// </summary>
+        int ScreenshotQuality { get; set; }
+
+        /// <summary>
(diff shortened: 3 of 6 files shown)


Runtime/BacktraceDatabase.cs
```diff
@@ -52,6 +52,37 @@ public IBacktraceBreadcrumbs Breadcrumbs
...
+            }
+            set
+            {
+                BacktraceDatabaseFileContext.ScreenshotQuality = value;
+            }
+        }
+
+        /// <summary>
+        /// Attachment support: Screenshot max height - based on screenshot max height, algorithm calculates
+        /// ration, that allows to calculate screenshot max width
```
rqbacktrace, line 72: Typo: ration -> ratio

Runtime/Interfaces/IBacktraceDatabaseFileContext.cs
```diff
@@ -7,6 +7,17 @@ namespace Backtrace.Unity.Interfaces
...
     internal interface IBacktraceDatabaseFileContext
     {
+        /// <summary>
+        /// Screenshot quality
+        /// </summary>
+        int ScreenshotQuality { get; set; }
+
+        /// <summary>
+        /// Screenshot max height - based on screenshot max height, algorithm calculates
+        /// ration, that allows to calculate screenshot max width
```
rqbacktrace, line 17: Typo: ration -> ratio

Runtime/Interfaces/IBacktraceDatabase.cs
```diff
@@ -14,6 +14,17 @@ namespace Backtrace.Unity.Interfaces
...
     public interface IBacktraceDatabase
     {
+        /// <summary>
+        /// Screenshot quality
+        /// </summary>
+        int ScreenshotQuality { get; set; }
+
+        /// <summary>
+        /// Screenshot max height - based on screenshot max height, algorithm calculates
+        /// ration, that allows to calculate screenshot max width
```
rqbacktrace, line 24: Typo: ration -> ratio

Runtime/Services/BacktraceDatabaseFileContext.cs
```diff
@@ -42,13 +43,58 @@ internal class BacktraceDatabaseFileContext : IBacktraceDatabaseFileContext
...
+                {
+                    throw new ArgumentException($"{nameof(value)} cannot be bigger than 100");
+                }
+                _attachmentManager.ScreenshotQuality = value;
+            }
+        }
+
+        /// <summary>
+        /// Screenshot max height - based on screenshot max height, algorithm calculates
+        /// ration, that allows to calculate screenshot max width
```
rqbacktrace, line 79: Typo: ration -> ratio

Runtime/Model/Database/BacktraceDatabaseAttachmentManager.cs
```diff
@@ -93,23 +95,60 @@ private string GetScreenshotPath(string dataPrefix)
...
+
+#if UNITY_2019_1_OR_NEWER
+                        RenderTexture.ReleaseTemporary(screenTexture);
+#else
+                        GameObject.Destroy(screenTexture);
+#endif
+                    }
+                    else
+                    {
+                        result = ScreenCapture.CaptureScreenshotAsTexture();
```
rqbacktrace, line 146: How do we ensure this is not called during the rendering process?

From https://docs.unity3d.com/ScriptReference/ScreenCapture.CaptureScreenshotAsTexture.html

"To get a reliable output from this method you must make sure it is called once the frame rendering has ended, and not during the rendering process. A simple way of ensuring this is to call it from a coroutine that yields on WaitForEndOfFrame. If you call this method during the rendering process you will get unpredictable and undefined results."

Runtime/Model/Database/BacktraceDatabaseAttachmentManager.cs
```diff
@@ -93,23 +95,60 @@ private string GetScreenshotPath(string dataPrefix)
...
                     {
-                        fs.Write(bytes, 0, bytes.Length);
+
+                        float ratio = Screen.width / Screen.height;
+                        int targetHeight = Mathf.Min(Screen.height, ScreenshotMaxHeight);
+                        int targetWidth = Mathf.RoundToInt(targetHeight * ratio);
+
+#if UNITY_2019_1_OR_NEWER
+                        RenderTexture screenTexture = RenderTexture.GetTemporary(Screen.width, Screen.height);
+                        ScreenCapture.CaptureScreenshotIntoRenderTexture(screenTexture);
```
rqbacktrace, line 108: Is it worth doing this asynchronously? https://docs.unity3d.com/ScriptReference/ScreenCapture.CaptureScreenshotIntoRenderTexture.html

Especially if it is a non-fatal exception, maybe we don't want to block important game threads. I imagine the downsampling of the image will add some extra processing time as well. It might be worth profiling to see if we're actually saving time with this during runtime.

rqbacktrace (commented):
No blocking changes, but maybe we need to profile the new screenshot algorithm to see if we're actually saving time?

Runtime/Model/Database/BacktraceDatabaseAttachmentManager.cs
konraddysput: The problem that I have with async in this solution is that we will need to write async callbacks to continue error submission. Overall our solution uses coroutines. Just after this database operation, we will yield a return from the Backtrace client to free the current frame.

Runtime/Interfaces/IBacktraceDatabaseFileContext.cs
konraddysput: fixed, thx!

Runtime/Interfaces/IBacktraceDatabase.cs
konraddysput: fixed, thx!

Runtime/Model/Database/BacktraceDatabaseAttachmentManager.cs
rqbacktrace: Do we save time here? It looks like we still capture the screenshot at full resolution and then take additional time to do downsampling. It looks like we maybe are adding time overall. Maybe we don't care but I'm only asking because of this user story:

"As a user, I would like to define attachments settings that allow the game to save resources/time when the game generates image attachment."

I understand we would save resources by saving downsampled images, but are we saving time? I think we are probably adding time

rqbacktrace: I guess we can save time because we are spending less time on the file write. But it might be worth doing a simple test to check if we are actually saving time overall including the downsampling algorithm. Otherwise I'm sure the functionality works and there's no issues there.

rqbacktrace: I guess we can save time because we are spending less time on the file write. But it might be worth doing a simple test to check if we are actually saving time overall including the downsampling algorithm. Otherwise I'm sure the functionality works and there's no issues there.

konraddysput: Do we save time? 
I don't think so:
(without image modifications) 
image

(without image modification)
image

Do we save anything at all? Yes - transfer, attachment size, disk usage.

rqbacktrace: Sounds good, if user enables this setting they will have smaller attachment sizes which can enable user to capture more screenshots and otherwise save on disk usage, as well as save transfer time at crash time. But user will not save during run time. Thank you for providing the numbers!

rqbacktrace (approved):
Looks good to me!

PR #92, opened 2021-06-07
Ignore LogType.Error

konraddysput:
## Why

As a user, I would like to ignore all log type error and prevent them from sending data to Backtrace. Usually LogType.Errors are not important from me from an analysis point of view. Somehow in more advanced examples or in specific game releases, I would like to enable them. As a user, I would like to use an option that allows me to define when we should send errors or not.


Runtime/BacktraceClient.cs
```diff
@@ -930,25 +930,34 @@ internal void HandleLowMemory()
...
-                // detect sampling flow
-                // we should apply sampling only to unhandled exceptions that are type LogType == Error
-                // log type error won't provide full exception information
-                if (type == LogType.Error && SamplingShouldSkip())
+                return;
+            }
+            BacktraceUnhandledException exception = null;
+            var invokeSkipApi = true;
+            // detect sampling flow
+            // we should apply sampling only to unhandled exceptions that are type LogType == Error
```
rqbacktrace, line 944: My understanding here is we only sample LogType.Error messages because they can be quite spammy. For other exception/error types we will always send and never do any sampling. And also we can use filtering to filter out certain kinds of reports.

Runtime/BacktraceClient.cs
```diff
@@ -930,25 +930,34 @@ internal void HandleLowMemory()
...
-                var invokeSkipApi = true;
-
-                // detect sampling flow
-                // we should apply sampling only to unhandled exceptions that are type LogType == Error
-                // log type error won't provide full exception information
-                if (type == LogType.Error && SamplingShouldSkip())
+                return;
+            }
+            BacktraceUnhandledException exception = null;
+            var invokeSkipApi = true;
```
rqbacktrace, line 942: Can this ever be false? Who can modify this guy?

rqbacktrace (approved):
Looks good to me!

Runtime/BacktraceClient.cs
konraddysput: Thats true - I think we should adjust the comment to be sure that it points to LogType.Error not to unhandled exceptions, right?

Runtime/BacktraceClient.cs
konraddysput: Yes - if Sampling hit and SkipReport callback said "no, we shouldn't filter this report". In this case we don't need to invoke SkipCallback anymore in the `ShouldSendReport` method.

PR #97, opened 2021-08-03
Windows native support

konraddysput:
# Windows native support

This diff allows us to collect native information on Windows builds. This diff also allows us to capture:
* ANRs 
* dynamic attributes
* collaborate with Unity crash reporting tool (in case when Unity crash reporting tool prevents from generating dump info)
* attachment support for native reports,

on Windows. Also this diff has small improvement for query parameter managements - previously we created query parameters in the submission url - now we're adding attributes to the post form data.


Runtime/Native/Windows/NativeClient.cs
```diff
@@ -0,0 +1,438 @@
...
+        public void HandleAnr(string gameObjectName = "", string callbackName = "")
+        {
+            if (!_captureNativeCrashes || !_configuration.HandleANR)
+            {
+                return;
+            }
+
+            bool reported = false;
+            var mainThreadId = Thread.CurrentThread.ManagedThreadId;
+            _anrThread = new Thread(() =>
```
vlussenburg, line 177: I recognize this code, sorry for being lazy but is it duplicated or just moved?

Runtime/Native/Windows/NativeClient.cs
```diff
@@ -0,0 +1,438 @@
...
+            foreach (var attributeKey in attributes.Keys)
+            {
+                var value = PlayerPrefs.GetString(string.Format(ScopedAttributesPattern, attributeKey), string.Empty);
+                result[attributeKey] = value;
+            }
+
+            // extend scoped attributes with legacy attributes stored by Backtrace-Unity library in previous versions
+            IDictionary<string, string> legacyAttributes = new Dictionary<string, string>()
+                    {
+                        { MachineUuidKey, "guid" },
```
vlussenburg, line 362: What are these legacy attributes?

Runtime/Native/Windows/NativeClient.cs
```diff
@@ -0,0 +1,438 @@
...
+                {
+                    PlayerPrefs.DeleteKey(legacyAttribute.Key);
+                    result[legacyAttribute.Value] = legacyAttributeValue;
+                }
+            }
+            return result;
+        }
+
+        /// <summary>
+        /// Adds attributes to scoped registry and to native clietn
```
vlussenburg, line 380: typo clietn

Runtime/Native/Windows/NativeClient.cs
```diff
@@ -0,0 +1,438 @@
...
+        {
+            if (_captureNativeCrashes)
+            {
+                AddNativeAttribute(key, value);
+            }
+            AddScopedAttribute(key, value);
+        }
+
+        /// <summary>
+        /// Adds dictionary of attributes to player prefs for windows crashes captured by unity crash handler
```
vlussenburg, line 394: Is this ONLY for unity crash handler or Crashpad too?

Runtime/Native/Windows/NativeClient.cs
```diff
@@ -0,0 +1,438 @@
...
+            foreach (var attribute in attributes)
+            {
+                attributesContainer.Keys.Add(attribute.Key);
+                PlayerPrefs.SetString(string.Format(ScopedAttributesPattern, attribute.Key), attribute.Value);
+            }
+            PlayerPrefs.SetString(ScopedAttributeListKey, JsonUtility.ToJson(attributesContainer));
+        }
+
+        /// <summary>
+        /// Adds attribute to player prefs for windows crashes captured by Unity crash handler
```
vlussenburg, line 413: crashpad too?

Runtime/Services/BacktraceApi.cs
```diff
@@ -102,7 +102,7 @@ public string ServerUrl
         /// <param name="attachments">List of attachments</param>
         /// <param name="callback">Callback</param>
         /// <returns>Server response</returns>
-        public IEnumerator SendMinidump(string minidumpPath, IEnumerable<string> attachments, IDictionary<string, string> queryAttributes, Action<BacktraceResult> callback = null)
```
vlussenburg, line 105: can moving from query to form attribute break anything? I guess not since it's POST and for minidump specifically?

StreamingAssets.meta
```diff
@@ -0,0 +1,8 @@
+fileFormatVersion: 2
```
vlussenburg, line 1: should this file be here? StreamingAssets?

vlussenburg (changes requested):

Runtime/Native/Windows/NativeClient.cs
konraddysput: we have dedicated ANR thread per native integration. it's the same integration that we have in other native clients and we should move our integration to generic class, however it's out of scope and generic implementation might break stuff on Android/iOS without additional tests.

Runtime/Native/Windows/NativeClient.cs
konraddysput: In 3.5.0 we added some attributes to PlayerPrefs (listed above) and now when we're migrating to full list of attributes we still want to be sure, we are able to upload them when we stored them in previous session. Legacy attributes exists now in case when a game crashed and after update (next launch) we should still send a minidump file with guid, session, machine id etc.

Runtime/Native/Windows/NativeClient.cs
konraddysput: Only Unity crash handler - Crashpad attributes are stored by different flow and inside the native client.

Runtime/Native/Windows/NativeClient.cs
konraddysput: Only Unity crash handler - Crashpad attributes are stored by different flow and inside the native client.

Runtime/Services/BacktraceApi.cs
konraddysput: No it wont - we're using the same pattern in different clients - example: https://github.com/backtrace-labs/backtrace-cocoa/blob/7dacd794a9318816121eafdb7d0a6e52a8cf8e55/Sources/Features/Network/MultipartRequest.swift#L55

StreamingAssets.meta
konraddysput: Good point - I will remove it, THanks!

vlussenburg (approved):

Runtime/Native/Windows/NativeClient.cs
vlussenburg: Ok. Can you please add it to technical debt for future work?

Runtime/Native/Windows/NativeClient.cs
vlussenburg: Ok. Should we remove this code in 3.7? Should be add a reminder to ourselves?

Runtime/Services/BacktraceApi.cs
vlussenburg: ok!

Runtime/Native/Windows/NativeClient.cs
konraddysput: No, because there will be always a chance that someone will migrate from 3.5 to n. I think we should remove it once we will introduce breaking change

Runtime/Native/Windows/NativeClient.cs
```diff
@@ -0,0 +1,438 @@
...
+
+                                reported = true;
+                                if (AndroidJNI.AttachCurrentThread() == 0)
+                                {
+                                    // set temporary attribute to "Hang"
+                                    AddNativeAttribute("error.type", "Hang");
+
+                                    NativeReport("ANRException: Blocked thread detected.", true);
+                                    // update error.type attribute in case when crash happen 
+                                    AddNativeAttribute("error.type", "Crash");
```
rqbacktrace, line 201: Is it possible that another crash might occur in another thread before we can set `error.type` back to `Crash`?

Runtime/Native/Windows/NativeClient.cs
```diff
@@ -0,0 +1,438 @@
...
+            if (!_captureNativeCrashes || !_configuration.HandleANR)
+            {
+                return;
+            }
+
+            bool reported = false;
+            var mainThreadId = Thread.CurrentThread.ManagedThreadId;
+            _anrThread = new Thread(() =>
+            {
+                float lastUpdatedCache = 0;
```
rqbacktrace, line 179: A name like `cachedLastUpdateTime` or `lastUpdateTimeCached` might be more clear

Runtime/Native/Windows/NativeClient.cs
```diff
@@ -0,0 +1,438 @@
...
+
+                                    NativeReport("ANRException: Blocked thread detected.", true);
+                                    // update error.type attribute in case when crash happen 
+                                    AddNativeAttribute("error.type", "Crash");
+                                }
+                            }
+                        }
+                        else
+                        {
+                            reported = false;
```
rqbacktrace, line 207: Won't this logic still just keep sending infinite ANR reports while `lastUpdatedCache == _lastUpdateTime`? Every iteration we will set `reported = true`, then next iteration we will again set `reported = false` and send another ANR report.

Runtime/Native/Windows/NativeClient.cs
```diff
@@ -0,0 +1,438 @@
...
+                yield break;
+            }
+            var crashDirs = Directory.GetDirectories(nativeCrashesDir);
+            foreach (var crashDir in crashDirs)
+            {
+
+                var crashDirFullPath = Path.Combine(nativeCrashesDir, crashDir);
+                var crashFiles = Directory.GetFiles(crashDirFullPath);
+
+                var alreadyUploaded = crashFiles.Any(n => n.EndsWith("backtrace.json"));
```
rqbacktrace, line 276: Does each minidump get stored in its own directory?

Runtime/Native/Windows/NativeClient.cs
```diff
@@ -0,0 +1,438 @@
...
+                }
+                if (attachments == null)
+                {
+                    attachments = new List<string>();
+                }
+                attachments.Concat(crashFiles.Where(n => n != minidumpPath));
+                IDictionary<string, string> attributes = GetScopedAttributes();
+                // be sure that error.type attribute provided by default by our library
+                // is always present in native attributes.
+                attributes["error.type"] = "Crash";
```
rqbacktrace, line 294: Should we check previous value of `attributes["error.type"]` so we don't overwrite something in case we have a Hang or some other kind of `error.type` here? Or are these only dumps generated by Unity in which case it's safe?

Runtime/Native/Windows/NativeClient.cs
```diff
@@ -0,0 +1,438 @@
...
+            return Path.Combine(Application.dataPath, pluginHandlerPath);
+
+        }
+        /// <summary>
+        /// Clean scoped attributes
+        /// </summary>
+        internal static void CleanScopedAttributes()
+        {
+            // cleaning scoped attributes should be skipped when 
+            // Configuration.SendUnhandledGameCrashesOnGameStartup  is set to false
```
rqbacktrace, line 328: I see we always `CleanScopedAttributes` when we construct `NativeClient`. I think we don't check `Configuration.SendUnhandledGameCrashesOnGameStartup` before we clean scoped attributes.

Runtime/Native/Windows/NativeClient.cs
```diff
@@ -0,0 +1,438 @@
...
+                        { VersionKey, "application.version" },
+                        { SessionKey, BacktraceMetrics.ApplicationSessionKey }
+                    };
+
+            foreach (var legacyAttribute in legacyAttributes)
+            {
+                string legacyAttributeValue = PlayerPrefs.GetString(legacyAttribute.Key, string.Empty);
+                if (!string.IsNullOrEmpty(legacyAttributeValue))
+                {
+                    PlayerPrefs.DeleteKey(legacyAttribute.Key);
```
rqbacktrace, line 372: Why should we delete legacy attributes from `PlayerPrefs`?

rqbacktrace (commented):
I don't see any need to block if the customer need is urgent but I have a few concerns about the `SendUnhandledGameCrashesOnGameStartup` Coroutine and the logic in `HandleANR`

Runtime/Model/BacktraceConfiguration.cs
```diff
@@ -120,10 +120,10 @@ public class BacktraceConfiguration : ScriptableObject
         /// Send Out of memory exceptions to Backtrace. 
         /// </summary>
         [Tooltip("Send Out of Memory exceptions to Backtrace")]
-#endif
         public bool OomReports = false;
+#endif
 
-#if UNITY_2019_2_OR_NEWER
+#if UNITY_2019_2_OR_NEWER && UNITY_ANDROID
```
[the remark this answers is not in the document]
konraddysput: This feature is available only on Android and only in Unity 2019+

Runtime/Services/BacktraceApi.cs
```diff
@@ -202,11 +198,7 @@ public IEnumerator Send(string json, IEnumerable<string> attachments, Dictionary
               ? System.Diagnostics.Stopwatch.StartNew()
               : new System.Diagnostics.Stopwatch();
 
-            var requestUrl = queryAttributes != null
-                ? GetParametrizedQuery(_serverUrl.ToString(), queryAttributes)
-                : ServerUrl;
-
-            using (var request = _httpClient.Post(requestUrl, json, attachments))
+            using (var request = _httpClient.Post(ServerUrl, json, attachments, queryAttributes))
```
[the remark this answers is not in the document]
konraddysput: query attributes are collected outside algorithm responsible for generating JSON data - for example to collect performance statistics: how much time we spend on generating report/data, storing data on the database etc - something that we can't just put in attributes. They suppose to in query attributes, but since I adjusted our API for minidump purpose they can be renamed to something else. Since you know the purpose of those attributes belong to current request or they collect statistics for current request - `requestAttributes` is a better name? What do you think?

Tests/Runtime/Native/Windows/ScopedNativeAttributesTests.cs
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        public void FreshStartup_ShouldntIncludeAnyAttributeFromPlayerPrefs_AttributesAreEmpty()
+        {
+            var attributes = NativeClient.GetScopedAttributes();
+
+            Assert.IsEmpty(attributes);
+        }
+
+        public void LegacyAttributesSupport_ShouldIncludeLegacyAttributesWhenScopedAttributesAreNotAvailable_AllLegacyAttributesArePresent()
```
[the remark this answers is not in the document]
konraddysput:
```
  string testVersion = "0.1.0";
            PlayerPrefs.SetString(NativeClient.VersionKey, testVersion);
            string machineUuid = "foo-bar";
            PlayerPrefs.SetString(NativeClient.MachineUuidKey, machineUuid);
            string sessionKey = "session-foo-bar-baz";
            PlayerPrefs.SetString(NativeClient.SessionKey, sessionKey);
```

this code is responsible for setting legacy attributes. Scoped attributes (that we added in this patch) are not available. `GetScopedAttributes` gets all attributes that belongs to scoped attributes (new attributes + legacy attributes). Scoped attributes in this test have only legacy attributes.

Runtime/BacktraceClient.cs
```diff
@@ -513,17 +514,23 @@ public void Refresh()
...
-
+            // send minidump files generated by unity engine or unity game, not captured by Windows native integration
+            // this integration should start before native integration and before breadcrumbs integration
+            // to allow algorithm to send breadcrumbs file - if the breadcrumb file is available
+            var scopedAttributes = AttributeProvider.GenerateAttributes(false);
             if (Configuration.SendUnhandledGameCrashesOnGameStartup && isActiveAndEnabled)
             {
-                var nativeCrashUplaoder = new NativeCrashUploader(AttributeProvider, BacktraceApi);
-                StartCoroutine(nativeCrashUplaoder.SendUnhandledGameCrashesOnGameStartup());
+                StartCoroutine(Runtime.Native.Windows.NativeClient.SendUnhandledGameCrashesOnGameStartup(nativeAttachments, BacktraceApi));
```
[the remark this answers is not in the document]
konraddysput: It won't but we're reading breadcrumbs in memory in SendUnhandledGameCrashesOnGameStartup method before we start yielding.

Runtime/Native/Windows/NativeClient.cs
konraddysput: yes it is.

Runtime/Native/Windows/NativeClient.cs
konraddysput: no - every minidump file in this directory is a crash.

Runtime/Native/Windows/NativeClient.cs
konraddysput: We're reading them to memory before we start yielding.

Runtime/Native/Windows/NativeClient.cs
konraddysput: Because this allows us to migrate legacy attributes from version 3.5.0 to new attributes system in 3.6.0 and use always standard flow. We don't need them anymore in playerprefs.

Runtime/Native/Windows/NativeClient.cs
konraddysput: yes, there is a small chance that this migh happen

Runtime/Native/Windows/NativeClient.cs
konraddysput: we won't enter this else statement when we're in the middle of ANR (lastUpdatedCache == _lastUpdateTime)

Runtime/Services/BacktraceApi.cs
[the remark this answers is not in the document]
rqbacktrace: I like `requestAttributes`!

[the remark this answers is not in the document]
rqbacktrace: @konraddysput what about `httpRequestAttributes` or `backtraceRequestAttributes` or `metaAttributes` or something?

Runtime/BacktraceClient.cs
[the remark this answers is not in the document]
rqbacktrace: In that case we would only be able to trust the breadcrumbs for the first iteration of the loop right? On next iteration (next minidump) we would have started breadcrumbs integration and erased old breadcrumbs already?

Runtime/Native/Windows/NativeClient.cs
rqbacktrace: Maybe it's worth it to add a comment to warn users of this potentially unexpected behavior?

Runtime/Native/Windows/NativeClient.cs
rqbacktrace: But ANR thread is a separate background thread running a while loop, it is not tied to game update logic. It is possible that `lastUpdatedCache == _lastUpdateTime` while `_lastUpdateTime` doesn't change for a long time, causing us iterate this loop for an indefinite number of times while game engine is not responsive.

Runtime/Native/Windows/NativeClient.cs
```diff
@@ -0,0 +1,438 @@
...
+        }
+        /// <summary>
+        /// Clean scoped attributes
+        /// </summary>
+        internal static void CleanScopedAttributes()
+        {
+            // cleaning scoped attributes should be skipped when 
+            // Configuration.SendUnhandledGameCrashesOnGameStartup  is set to false
+            // the reason behind this decision is to make sure user change in the configuration
+            // won't left and useless data
```
rqbacktrace, line 330: Did you mean `won't leave any useless data`?

konraddysput: Yeah, I meant that. Thanks!

Runtime/BacktraceClient.cs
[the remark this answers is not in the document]
konraddysput: This is a really good point - we need to make a copy of the breadcrumb file and use a copy instead of source breadcrumbs file.

Runtime/Services/BacktraceApi.cs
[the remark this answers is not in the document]
konraddysput: I just renamed it to `attributes`. Thanks for suggestion

Runtime/Native/Windows/NativeClient.cs
konraddysput: > But ANR thread is a separate background thread running a while loop, it is not tied to game update logic. 
It's not true. Background thread shares a reference to _lastUpdateTime with main thread. Every time when main thread updates _lastUpdateTime, the background thread is able to detect it and then set it to `false`.

```
 It is possible that lastUpdatedCache == _lastUpdateTime while _lastUpdateTime doesn't change for a long time, causing us iterate this loop for an indefinite number of times while game engine is not responsive.
```
Based on what I see and how it works it's impossible for me - I don't see a way how it's possible.

konraddysput:
@rqbacktrace Adjusted `SendUnhandledGameCrashesOnGameStartup` - now it will use a temporary copy of the breadcrumb file.

Runtime/Native/Windows/NativeClient.cs
```diff
@@ -0,0 +1,438 @@
...
+
+            var minidumpUrl = new BacktraceCredentials(_configuration.GetValidServerUrl()).GetMinidumpSubmissionUrl().ToString();
+
+            if (!Directory.Exists(databasePath))
+            {
+                Directory.CreateDirectory(databasePath);
+            }
+
+            // reassign to captureNativeCrashes
+            // to avoid doing anything on crashpad binary, when crashpad isn't available
```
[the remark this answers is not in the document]
rqbacktrace: Sure if it matches the configuration option name it makes sense, thank you!

Runtime/Native/Windows/NativeClient.cs
rqbacktrace: Look at this loop, it will iterate as fast as possible, this loop is not bound to the Unity game `Update` method:
`                while (_anrThread.IsAlive && _stopAnr == false)`

This means that even if `lastUpdatedCache == _lastUpdateTime` only for one Unity update frame (8 - 16 ms depending on frame rate), the while loop will still go into this path many times, potentially sending multiple reports:
`                        else if (lastUpdatedCache == _lastUpdateTime)`

Does that make sense? Because while loop will try to execute as fast as possible it can happen many times even within one frame

Runtime/Native/Windows/NativeClient.cs
```diff
@@ -0,0 +1,486 @@
...
+                    {
+                        Debug.LogWarning(string.Format("Cannot make a copy of the breadcrumb file in the database directory. Reason: {0}", e.Message));
+                    }
+                    finally
+                    {
+                        requireBreadcrumbsCopy = false;
+                    }
+                }
+                var dumpAttachment = crashFiles.Concat(attachments).Where(n => n != minidumpPath).ToList();
+                IDictionary<string, string> attributes = GetScopedAttributes();
```
rqbacktrace, line 326: Should we cache the `GetScopedAttributes` too like we do with Breadcrumbs file now?

Runtime/Native/Windows/NativeClient.cs
rqbacktrace: I think we need to do the same thing we do with Breadcrumbs file now

Runtime/Native/Windows/NativeClient.cs
konraddysput: Yes - we should. Thanks for pointing this out!

rqbacktrace (approved):
Looks good to me, thank you!

PR #98, opened 2021-08-03
Source Code integration

konraddysput:
# Why 

This diff brings back log buffer to Unity reports (source code integration)


Runtime/Model/BacktraceReport.cs
```diff
@@ -125,6 +130,24 @@ private void SetDefaultAttributes()
...
+            {
+                return;
+            }
+
+            SourceCode = new BacktraceSourceCode()
+            {
+                Text = text
+            };
+            // assign log information to first stack frame
+            DiagnosticStack[0].SourceCode = BacktraceSourceCode.SOURCE_CODE_PROPERTY;
```
vlussenburg, line 149: @konraddysput can you add the source code to EACH FRAME of the faulting thread? I know there's a ticket for this because lots of users like to have the logs for whatever frame they click on.

I can't find the ticket RN but I am sure it was customer feedback.

vlussenburg (approved):

Runtime/Model/BacktraceUnityMessage.cs
```diff
@@ -0,0 +1,93 @@
...
+        private readonly string _formattedMessage;
+        public readonly string Message;
+        public readonly string StackTrace;
+        public readonly LogType Type;
+
+        public BacktraceUnityMessage(BacktraceReport report)
+        {
+            if (report == null)
+            {
+                throw new ArgumentException("report");
```
rqbacktrace, line 22: BacktraceUnityMessage constructor can throw but we don't do exception handling when we construct BacktraceUnityMessage throughout the code

Runtime/Model/BacktraceUnityMessage.cs
```diff
@@ -0,0 +1,93 @@
...
+        /// </summary>
+        /// <returns>True if unity message is an exception/error message</returns>
+        public bool IsUnhandledException()
+        {
+            return ((Type == LogType.Exception || Type == LogType.Error)
+                && !string.IsNullOrEmpty(Message));
+        }
+
+        /// <summary>
+        /// Convert Backtrace Untiy Message to string that will be acceptable by source code format
```
rqbacktrace, line 85: `Untiy` -> `Unity`

rqbacktrace (commented):
No need to block but maybe it would be safer to make sure we handle potential exception if we fail to construct `BacktraceUnityMessage`?

Runtime/Model/BacktraceUnityMessage.cs
konraddysput: I think it's expected situation - the reason behind this decision is to let the developer know about mistakes when the developer creates new functionality around BacktraceUnityMessage. Our code will never throw this exception - all reports are defined and there won't be an option in our API that the report object will be undefined. 

This conditions check future improvements to make sure we have valid information required to create a Unity message that we can change into the specific log.

Runtime/Model/BacktraceUnityMessage.cs
konraddysput: fixed, thx

Runtime/Model/BacktraceUnityMessage.cs
rqbacktrace: Sounds good, thank you!

rqbacktrace (approved):
Looks good to me!

Runtime/Model/BacktraceReport.cs
konraddysput: It's now available for every frame.

PR #109, opened 2021-09-27
Escape control characters

konraddysput:
# Why

We detected a problem with the control characters that the Unity plugin sends. Without control characters escaping mechanism, invalid JSON provided by Unity plugin might break web ui parser. This diff correctly escapes control characters in the plugin.


rqbacktrace (approved):
Looks good! I didn't see anything obvious and it's a little hard to validate the encoding/decoding logic but I trust your unit test. Paul would probably be best to verify the logical correctness here.

vlussenburg (approved):

PR #110, opened 2021-09-28
Backtrace thread API improvements

konraddysput:
# Why

This diff contains safe code that we should execute when a game executes Backtrace client code outside the main thread. This code also solves problems detected while I was trying to figure our correct solution to this issue.

### DIFF
--- a/Runtime/BacktraceDatabase.cs
+++ b/Runtime/BacktraceDatabase.cs
@@ -581,7 +581,7 @@ namespace Backtrace.Unity
             DatabasePath = Configuration.GetFullDatabasePath();
             if (string.IsNullOrEmpty(DatabasePath))
             {
-                Debug.LogWarning("Backtrace database path is empty or unvailable.");
+                Debug.LogWarning("Backtrace database path is empty or unavailable.");
                 return false;
             }
 
--- a/Runtime/BacktraceDatabase.cs
+++ b/Runtime/BacktraceDatabase.cs
@@ -581,7 +581,7 @@ namespace Backtrace.Unity
             DatabasePath = Configuration.GetFullDatabasePath();
             if (string.IsNullOrEmpty(DatabasePath))
             {
-                Debug.LogWarning("Backtrace database path is empty.");
+                Debug.LogWarning("Backtrace database path is empty or unvailable.");
                 return false;
             }
 
--- a/Runtime/Common/ClientPathHelper.cs
+++ b/Runtime/Common/ClientPathHelper.cs
@@ -59,7 +59,14 @@ namespace Backtrace.Unity.Common
             {
                 path = Path.Combine(Application.persistentDataPath, path);
             }
-            return Path.GetFullPath(path);
+            try
+            {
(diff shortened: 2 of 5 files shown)


Runtime/BacktraceDatabase.cs
```diff
@@ -581,7 +581,7 @@ protected virtual bool InitializeDatabasePaths()
             DatabasePath = Configuration.GetFullDatabasePath();
             if (string.IsNullOrEmpty(DatabasePath))
             {
-                Debug.LogWarning("Backtrace database path is empty.");
+                Debug.LogWarning("Backtrace database path is empty or unvailable.");
```
rqbacktrace, line 584: Typo: `unvailable` -> `unavailable`

rqbacktrace (approved):
I see we are using `System.Random` as Unity docs suggest we should do if working with threads, and overall increasing error handling.

Looks good with just one minor typo!

Runtime/BacktraceDatabase.cs
vlussenburg: +1

Runtime/Model/Database/BacktraceDatabaseRecord.cs
```diff
@@ -181,6 +181,10 @@ public virtual void Increment()
         /// <returns>Saved database record</returns>
         internal static BacktraceDatabaseRecord ReadFromFile(FileInfo file)
         {
+            if (!file.Exists)
```
vlussenburg, line 184: Null check?

Runtime/Common/ClientPathHelper.cs
```diff
@@ -59,7 +59,14 @@ private static string GenerateFullPath(this string path)
...
                 path = Path.Combine(Application.persistentDataPath, path);
             }
-            return Path.GetFullPath(path);
+            try
+            {
+                return Path.GetFullPath(path);
+            }
+            catch (Exception)
+            {
+                return string.Empty;
```
vlussenburg, line 68: Would add a debug log message so the Exception is logged somewhere. Now, if this code has a problem, the exception details are completely swallowed and can't be retrieved in any way.

vlussenburg (changes requested):

Runtime/Common/ClientPathHelper.cs
konraddysput: It is in the higher level - it doesn't make sense to include debug logs in the helper, since later we will have a problem with helpers dependencies.

Runtime/Model/Database/BacktraceDatabaseRecord.cs
konraddysput: This is an internal method that our api call. We already know in higher level that file is defined.

```
   var files = BacktraceDatabaseFileContext.GetRecords();
            foreach (var file in files)
            {
                var record = BacktraceDatabaseRecord.ReadFromFile(file);
```

vlussenburg (approved):

PR #111, opened 2021-09-28
Multiselect dropdown support

konraddysput:
# Why

We discovered that the `EditorGUILayout.PropertyField` API that the Unity plugin uses to draw Backtrace configuration options on the UI doesn't work correctly on the old version of Unity with enum flags. By default, the user instead of selecting multiple values can select only one. 

This diff allows user to select multiple values in the backtrace-unity configuration window in the Unity 2017/2018

### DIFF
--- a/Editor/BacktraceConfigurationEditor.cs
+++ b/Editor/BacktraceConfigurationEditor.cs
@@ -221,7 +221,7 @@ namespace Backtrace.Unity.Editor
         /// if editor has to display multiselect dropdown by using enum flags. This code allows to generate
         /// multiselect dropdown based on the available enum.
         /// </summary>
-        /// <param name="propertyName">Enum proberty name</param>
+        /// <param name="propertyName">Enum property name</param>
         /// <param name="label">Label</param>
         /// <param name="serializedObject">Serialized object</param>
         private static void DrawMultiselectDropdown(string propertyName, string label, SerializedObject serializedObject)
--- a/Editor/BacktraceConfigurationEditor.cs
+++ b/Editor/BacktraceConfigurationEditor.cs
@@ -1,4 +1,5 @@
 using Backtrace.Unity.Model;
+using Backtrace.Unity.Model.Breadcrumbs;
 using System;
 using UnityEditor;
 using UnityEngine;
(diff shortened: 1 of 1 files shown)


Editor/BacktraceConfigurationEditor.cs
```diff
@@ -225,5 +215,29 @@ public override void OnInspectorGUI()
 
             serializedObject.ApplyModifiedProperties();
         }
+
+        /// <summary>
+        /// Draw multiselect dropdown. By default PropertyField won't work correctly in Untiy 2017/2018
```
rqbacktrace, line 220: Typo: `Untiy` -> `Unity`

Editor/BacktraceConfigurationEditor.cs
```diff
@@ -225,5 +215,29 @@ public override void OnInspectorGUI()
...
+        /// if editor has to display multiselect dropdown by using enum flags. This code allows to generate
+        /// multiselect dropdown based on the available enum.
+        /// </summary>
+        /// <param name="propertyName">Enum proberty name</param>
```
rqbacktrace, line 224: Typo: `proberty` -> `property`

Editor/BacktraceConfigurationEditor.cs
```diff
@@ -1,4 +1,5 @@
 using Backtrace.Unity.Model;
+using Backtrace.Unity.Model.Breadcrumbs;
```
rqbacktrace, line 2: Why do we need `Breadcrumbs` include now?

rqbacktrace (approved):
Looks good! Just a few typos and question about `Breadcrumbs` include but I will go ahead and approve!

Editor/BacktraceConfigurationEditor.cs
konraddysput: This change needs to analyse possible labels available in multi select dropdown by using reflection mechanism on enum type. Since we're displaying options for enum we need to include this using to start using it in the reflection code.

vlussenburg (approved):

Editor/BacktraceConfigurationEditor.cs
rqbacktrace: Cool makes sense, thank you!

PR #112, opened 2021-09-30
Prevent from changing source code collection while jobject iterate ov…

konraddysput:
# Why

DOTS parallel job showed a problem with source code integration. When multiple jobs generate backtrace report, BacktraceJObject access LogQueue to generate unity engine logs, while other jobs might generate logs and modify enumerator used by BacktraceJobject.


rqbacktrace (approved):
Looks good to me! I see we create a copy of the logs before we use them so if they get modified by other jobs/threads it won't affect us.

PR #113, opened 2021-09-30
Bugfix/flag overflow

konraddysput:
# Why

Backtrace-Unity minidump options allow to set up a flag that will cause an int overflow. This diff changes the value type in BacktraceConfiguration property to accept long instead of int.


rqbacktrace (approved):
Looks good to me!

PR #114, opened 2021-09-30
Release 3.6.2

konraddysput:
# Backtrace-Unity release 3.6.2. 

Changelog 

- Fixed invalid behavior of multi select dropdown in Unity 2017/2018. Now users can select multiple values (instead of one) on multi select dropdowns. 
- Fixed invalid escaping of control characters,
- Fixed a bug caused by Unity API when Backtrace tries to parse Unity message from a background thread.
- General API improvements


vlussenburg (commented):
LGTM. Let's sync on if we want to add other things to the scope or cut the release.

vlussenburg (approved):

PR #125, opened 2021-10-21
Remove Android dependencies on non-Android platforms

konraddysput:
# Why

This diff removes an Android dependency on the Windows platform.


konraddysput:
> Looks good to me! Only question I have is: Can we conditionally include `Android/NativeClient.cs`, `Windows/NativeClient.cs` etc. instead of conditionally commenting out the body of those files?

We were doing it, but if someone doesn't have an Android dependency in their source code, this will cause compilation issues.

PR #128, opened 2021-11-03
Breadcrumbs initialization improvements

konraddysput:
# Why

This diff contains small breadcrumbs initialization improvements. With these improvements, backtrace-unity will:
* disable breadcrumbs integration, when a client decides to not select any supported log level or any breadcrumb type,
* ignores files that are empty.

### DIFF
--- a/Runtime/BacktraceDatabase.cs
+++ b/Runtime/BacktraceDatabase.cs
@@ -36,7 +36,7 @@ namespace Backtrace.Unity
             {
                 if (_breadcrumbs == null)
                 {
-                    if (Enable && BacktraceBreadcrumbs.CanStoreBreadcrumbs(Configuration.LogLevel, Configuration.BacktraceBreadcrumbsLevel))
+                    if (Enable && Configuration.EnableBreadcrumbsSupport && BacktraceBreadcrumbs.CanStoreBreadcrumbs(Configuration.LogLevel, Configuration.BacktraceBreadcrumbsLevel))
                     {
                         _breadcrumbs = new BacktraceBreadcrumbs(new BacktraceStorageLogManager(Configuration.GetFullDatabasePath()));
                     }
--- a/Runtime/BacktraceDatabase.cs
+++ b/Runtime/BacktraceDatabase.cs
@@ -36,7 +36,7 @@ namespace Backtrace.Unity
             {
                 if (_breadcrumbs == null)
                 {
-                    if (Enable && Configuration.EnableBreadcrumbsSupport)
+                    if (Enable && BacktraceBreadcrumbs.CanStoreBreadcrumbs(Configuration.LogLevel, Configuration.BacktraceBreadcrumbsLevel))
                     {
                         _breadcrumbs = new BacktraceBreadcrumbs(new BacktraceStorageLogManager(Configuration.GetFullDatabasePath()));
                     }
--- a/Runtime/Model/BacktraceHttpClient.cs
+++ b/Runtime/Model/BacktraceHttpClient.cs
@@ -167,7 +167,7 @@ namespace Backtrace.Unity.Model
 
             foreach (var file in uniqueAttachments)
             {
-                if (string.IsNullOrEmpty(file) || File.Exists(file) == false)
+                if (string.IsNullOrEmpty(file) || !File.Exists(file))
                 {
                     continue;
                 }
--- a/Runtime/Model/BacktraceHttpClient.cs
+++ b/Runtime/Model/BacktraceHttpClient.cs
@@ -158,8 +158,8 @@ namespace Backtrace.Unity.Model
             {
                 return;
             }
-            // make sure attachments are not bigger than 10 Mb.
-            const int maximumAttachmentSize = 10000000;
+            // make sure attachments are not bigger than 10 MiB.
+            const int maximumAttachmentSize = 10 * 1024 * 1024;
             const string attachmentPrefix = "attachment_";
 
             var uniqueAttachments = new HashSet<string>(attachments.Reverse());
--- a/Runtime/Model/BacktraceHttpClient.cs
+++ b/Runtime/Model/BacktraceHttpClient.cs
@@ -167,7 +167,14 @@ namespace Backtrace.Unity.Model
 
             foreach (var file in uniqueAttachments)
             {
-                if (string.IsNullOrEmpty(file) || File.Exists(file) == false || new FileInfo(file).Length > maximumAttachmentSize)
+                if (string.IsNullOrEmpty(file) || File.Exists(file) == false)
+                {
+                    continue;
+                }
+
+                var fileSize = new FileInfo(file).Length;
+                var isInvalidFileSize = fileSize > maximumAttachmentSize || fileSize == 0;
+                if (isInvalidFileSize)
                 {
                     continue;
                 }
--- a/Runtime/Model/Breadcrumbs/BacktraceBreadcrumbs.cs
+++ b/Runtime/Model/Breadcrumbs/BacktraceBreadcrumbs.cs
@@ -185,7 +185,7 @@ namespace Backtrace.Unity.Model.Breadcrumbs
 
         public static bool CanStoreBreadcrumbs(UnityEngineLogLevel logLevel, BacktraceBreadcrumbType backtraceBreadcrumbsLevel)
         {
-            return (ulong)backtraceBreadcrumbsLevel != 0 && (ulong)logLevel != 0;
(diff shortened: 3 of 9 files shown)


Runtime/Model/BacktraceHttpClient.cs
```diff
@@ -167,7 +167,14 @@ private void AddAttachmentToFormData(List<IMultipartFormSection> formData, IEnum
...
             foreach (var file in uniqueAttachments)
             {
-                if (string.IsNullOrEmpty(file) || File.Exists(file) == false || new FileInfo(file).Length > maximumAttachmentSize)
+                if (string.IsNullOrEmpty(file) || File.Exists(file) == false)
+                {
+                    continue;
+                }
+
+                var fileSize = new FileInfo(file).Length;
+                var isInvalidFileSize = fileSize > maximumAttachmentSize || fileSize == 0;
```
rqbacktrace, line 176: Maybe it's worth separately logging if `fileSize > maximumAttachmentSize`?

Runtime/BacktraceDatabase.cs
```diff
@@ -36,7 +36,7 @@ public IBacktraceBreadcrumbs Breadcrumbs
             {
                 if (_breadcrumbs == null)
                 {
-                    if (Enable && Configuration.EnableBreadcrumbsSupport)
```
rqbacktrace, line 39: Is `EnableBreadcrumbsSupport` used anywhere else? Can we eliminate this variable from `Configuration` now and always use `BacktraceBreadcrumbs.CanStoreBreadcrumbs`?

Runtime/Model/Database/BacktraceDatabaseRecord.cs
```diff
@@ -95,7 +95,8 @@ public string BacktraceDataJson()
                 return null;
             }
 
-            return File.ReadAllText(DiagnosticDataPath);
+            DiagnosticDataJson = File.ReadAllText(DiagnosticDataPath);
+            return DiagnosticDataJson;
```
rqbacktrace, line 99: What's the reasoning behind this change?

rqbacktrace (approved):
Looks good to me with minor comments/questions, no blockers!

Runtime/Model/BacktraceHttpClient.cs
konraddysput: We didn't change this logic in this pull request ( we just added a condition to check if fileSize is equal to zero. Not sure if this is a good idea because we can spam too many logs when the game will generate unhandled exceptions in the update loop.

Runtime/BacktraceDatabase.cs
konraddysput: `CanStoreBreadcrumbs` is a method. `EnableBreadcrumbsSupport` is a configuration flag, that is responsible for setting breadcrumbs options at all.

Runtime/Model/Database/BacktraceDatabaseRecord.cs
konraddysput: On the top of this function, backtrace-unity checks if the record already cached JSON data. if it does, we return it. Otherwise, we're reading data. The problem that this change solves, allows us to cache reports even when we read it from the hard drive. This simple change can save time when someone reads data stored in the database.

Runtime/Model/BacktraceHttpClient.cs
rqbacktrace: Ah okay, I just brought it up because it may confuse the user when they are missing some file attachment, but I guess this condition `fileSize > maximumAttachmentSize` occurs mostly when there are unhandled exceptions in update loop so it won't happen often.

Runtime/BacktraceDatabase.cs
rqbacktrace: That makes sense, thank you!

konraddysput: Anyway, I didn't resolve the thread beacuse you were right! We should use here EnablebreadcrumbSupport flag anyway. Thanks for this catch @rqbacktrace !

Runtime/Model/Database/BacktraceDatabaseRecord.cs
rqbacktrace: Ahhh I see, that's pretty cool!

Runtime/BacktraceDatabase.cs
```diff
@@ -36,7 +36,7 @@ public IBacktraceBreadcrumbs Breadcrumbs
             {
                 if (_breadcrumbs == null)
                 {
-                    if (Enable && BacktraceBreadcrumbs.CanStoreBreadcrumbs(Configuration.LogLevel, Configuration.BacktraceBreadcrumbsLevel))
+                    if (Enable && Configuration.EnableBreadcrumbsSupport && BacktraceBreadcrumbs.CanStoreBreadcrumbs(Configuration.LogLevel, Configuration.BacktraceBreadcrumbsLevel))
```
rqbacktrace, line 39: 👍

Runtime/Model/BacktraceHttpClient.cs
```diff
@@ -167,7 +167,14 @@ private void AddAttachmentToFormData(List<IMultipartFormSection> formData, IEnum
 
             foreach (var file in uniqueAttachments)
             {
-                if (string.IsNullOrEmpty(file) || File.Exists(file) == false || new FileInfo(file).Length > maximumAttachmentSize)
+                if (string.IsNullOrEmpty(file) || File.Exists(file) == false)
+                {
+                    continue;
+                }
+
+                var fileSize = new FileInfo(file).Length;
```
perf2711, line 175: This is a potential race condition - the file could be deleted after `File.Exists()` returns `true`. As improbable as it is, this is possible.

Runtime/Model/BacktraceHttpClient.cs
```diff
@@ -167,7 +167,14 @@ private void AddAttachmentToFormData(List<IMultipartFormSection> formData, IEnum
 
             foreach (var file in uniqueAttachments)
             {
-                if (string.IsNullOrEmpty(file) || File.Exists(file) == false || new FileInfo(file).Length > maximumAttachmentSize)
+                if (string.IsNullOrEmpty(file) || File.Exists(file) == false)
```
perf2711, line 170: Why not `!File.Exists(file)`?

Runtime/Model/Breadcrumbs/BacktraceBreadcrumbs.cs
```diff
@@ -182,5 +182,10 @@ public void Update()
         {
             EventHandler.Update();
         }
+
+        public static bool CanStoreBreadcrumbs(UnityEngineLogLevel logLevel, BacktraceBreadcrumbType backtraceBreadcrumbsLevel)
+        {
+            return (ulong)backtraceBreadcrumbsLevel != 0 && (ulong)logLevel != 0;
```
perf2711, line 188: The enums could be extended by adding a `None` value, which equals to 0. Disabling the logging would be easier as well, no need to assign a `(UnityEngineLogLevel)0` casted value.

Runtime/Model/BacktraceHttpClient.cs
konraddysput: @perf2711 we need to get a file size - we don't want to check if a file exists. The file might exist, but the file size might be 0.

Runtime/Model/BacktraceHttpClient.cs
konraddysput: It doesn't matter, right?

Runtime/Model/Breadcrumbs/BacktraceBreadcrumbs.cs
konraddysput: If you want to extend an Enum with None value that is equal to zero, it might be a bad decision, since None is interpreted by Unity engine UI

konraddysput: At the beginning, I thought it won't work, but it looks like Unity handles it correctly. I'm validating it now with older Unity versions.

perf2711: All right.

Runtime/Model/BacktraceHttpClient.cs
perf2711: Sure.

Runtime/Model/BacktraceHttpClient.cs
perf2711: All right, I was just worried that `new FileInfo` may throw an exception, but it should not.

perf2711 (approved):

PR #130, opened 2021-11-04
Breadcrumbs ANR support

konraddysput:
# Why

This diff adds support for breadcrumbs log when ANR occurs. This diff also abstracts some logic from each native client and changes the way how BacktraceClient can initialize breadcrumbs integration. We need to know available options before we start native integration, so this requirement required us to rethink the way how we enable/start breadcrumbs integration.

### DIFF
--- a/Runtime/BacktraceClient.cs
+++ b/Runtime/BacktraceClient.cs
@@ -536,7 +536,7 @@ namespace Backtrace.Unity
                 Database = GetComponent<BacktraceDatabase>();
                 if (Database != null)
                 {
-                    _breadcrumbs = Database.Breadcrumbs as BacktraceBreadcrumbs;
+                    _breadcrumbs = (BacktraceBreadcrumbs)Database.Breadcrumbs;
                     Database.Reload();
                     Database.SetApi(BacktraceApi);
                     Database.SetReportWatcher(_reportLimitWatcher);
--- a/Runtime/BacktraceClient.cs
+++ b/Runtime/BacktraceClient.cs
@@ -25,8 +25,13 @@ namespace Backtrace.Unity
     public class BacktraceClient : MonoBehaviour, IBacktraceClient
     {
         public const string VERSION = "3.7.0-preview.2";
-
+        internal const string DefaultBacktraceGameObjectName = "BacktraceClient";
         public BacktraceConfiguration Configuration;
+
+        /// <summary>
+        /// Breadcrumbs instance for internal use. This instance allows to make system calls
+        /// </summary>
+        private BacktraceBreadcrumbs _breadcrumbs;
 
         /// <summary>
         /// Backtrace Breadcrumbs
@@ -35,11 +40,7 @@ namespace Backtrace.Unity
         {
             get
             {
-                if (Database == null)
-                {
-                    return null;
-                }
-                return Database.Breadcrumbs;
+                return _breadcrumbs;
             }
         }
 
@@ -383,7 +384,7 @@ namespace Backtrace.Unity
         /// param name="attachments">List of attachments </param>
         /// <param name="gameObjectName">game object name</param>
         /// <returns>Backtrace client</returns>
-        public static BacktraceClient Initialize(BacktraceConfiguration configuration, Dictionary<string, string> attributes = null, string gameObjectName = "BacktraceClient")
+        public static BacktraceClient Initialize(BacktraceConfiguration configuration, Dictionary<string, string> attributes = null, string gameObjectName = DefaultBacktraceGameObjectName)
         {
             if (string.IsNullOrEmpty(gameObjectName))
             {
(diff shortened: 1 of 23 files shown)


Runtime/BacktraceClient.cs
```diff
@@ -535,12 +536,13 @@ public void Refresh()
                 Database = GetComponent<BacktraceDatabase>();
                 if (Database != null)
                 {
+                    _breadcrumbs = Database.Breadcrumbs as BacktraceBreadcrumbs;
```
konraddysput, line 539: we can't provide to our clients with a real implementation (we hide it behind the interface) of the BacktraceBreadcrumbs class. Because of that, we weren't able to log correctly system events when a user didn't add "manual" options in the breadcrumbs configuration (because we use flow that is dedicated to users and we check if user has enabled manual logs). Since we own this implementation and we always create BacktraceBreadcrumbs class instance casting here is safe. If you have other suggestions how we can handle it, please let me know. We also need to share reference between two classes.

vlussenburg (approved):
LGTM. Needs Changelog which I will add.

Runtime/Native/Base/NativeClientBase.cs
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        /// <summary>
+        /// Update native client state
+        /// </summary>
+        /// <param name="time">Current game time</param>
+        public void Update(float time)
+        {
+            LastUpdateTime = time;
+            lock (_lockObject)
```
perf2711, line 69: If I understand correctly, this method is run basically every frame. Locks aren't known to be exactly cheap - also, if there is an event in which the lock is awaited to be freed, wouldn't the game freeze?

Runtime/BacktraceClient.cs
perf2711: Small nit: if casting here is safe, why not direct cast it: `(BacktraceBreadcrubms)Database.Breadcrumbs`? That way, if something changes in the future (different implementation), it should fail here instantly and not silently.

Runtime/Native/Base/NativeClientBase.cs
konraddysput: The reason behind it is to be sure we don't log something twice. a simple boolean flag might not be enough. This method might be invoked at the same time by more than 1 frame.  async/await is also unavailable here since it requires adding an additional package and maintaining it.  I thought lock is the cheapest solution that we have available.

perf2711: At least don't wait for the lock if you know nothing would happen.

```csharp
if (LogAnr) 
{
    if (Monitor.TryLock(_lockObject)) 
    {
        try 
        {
            if (LogAnr) 
            {
                // Push report
                LogAnr = false;
            }
        }
        finally
        {
            Monitor.Exit(_lockObject);
        }
    }
}
```

Tests/Runtime/Native/BreadcrumbsAnrTests.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            nativeClient.SimulateAnr();
+            nativeClient.Update(0);
+
+            Assert.AreEqual(expectedNumberOfBreadcrumbs, breadcrumbs.LogManager.Length());
+            Assert.AreEqual(TestableNativeClient.AnrMessage, backtraceStorageManager.Breadcrumbs.First().Message);
+        }
+
+        [Test]
+        public void TestAnrBreadcrumbReporting_ShouldAddSingleAnrBreadcrumb_SingleBreadcrumbSuccessfullyStored()
```
rqbacktrace, line 30: This test looks exactly the same as the previous test, do we need both?

rqbacktrace (approved):
Looks good to me with minor feedback/questions. Were you able to test this on all platforms which were affected by this change?

Runtime/Model/Breadcrumbs/IBacktraceBreadcrumbs.cs
```diff
@@ -7,6 +7,8 @@ namespace Backtrace.Unity.Model.Breadcrumbs
     public interface IBacktraceBreadcrumbs
     {
         BacktraceBreadcrumbType BreadcrumbsLevel { get; }
+        bool EnableBreadcrumbs();
+        [Obsolete("Please use EnableBreadcurmbs instead. This function will be removed in the future updates")]
```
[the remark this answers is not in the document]
konraddysput: done, thx!

Tests/Runtime/Native/BreadcrumbsAnrTests.cs
konraddysput: This test will invoke an update method multiple times. The test gonna validate if we will add breadcrumbs twice in a synchronous method.

perf2711 (approved):

PR #133, opened 2021-11-17
Apply default when value is below limit

konraddysput:
# Why

This diff adds support for default values in the UI. Previously a user knew that when the user launched the game. Now, we gonna apply some default every time when a user sets an invalid value. This pattern reuses the Unity editor pattern for unsigned integers that are available on the UI. Since this is the public API to make sure it won't break, we didn't change the variable types.

### DIFF
--- a/Runtime/Model/BacktraceConfiguration.cs
+++ b/Runtime/Model/BacktraceConfiguration.cs
@@ -28,6 +28,14 @@ namespace Backtrace.Unity.Model
             UnityEngineLogLevel.Info |
             UnityEngineLogLevel.Warning;
 
+        public const int DefaultRetryLimit = 3;
+        public const int DefaultReportPerMin = 50;
+        public const int DefaultGameObjectDepth = -1;
+        public const int DefaultNumberOfLogs = 10;
+        public const int DefaultMaxRecordCount = 8;
+        public const int DefaultMaxDatabaseSize = 0;
+        public const int DefaultRetryInterval = 60;
+
         /// <summary>
         /// Backtrace server url
         /// </summary>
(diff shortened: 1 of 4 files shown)


Runtime/Model/Database/BacktraceDatabaseSettings.cs
```diff
@@ -18,6 +19,7 @@ public BacktraceDatabaseSettings(string databasePath, BacktraceConfiguration con
 
             DatabasePath = databasePath;
             _configuration = configuration;
+            _retryInterval = configuration.RetryInterval >= 0 ? Convert.ToUInt32(_configuration.RetryInterval) : 60;
```
perf2711, line 22: Why not use a direct cast? Also, you did not use the constant defined in `BacktraceConfiguration`.

```csharp
_retryInterval = (uint)(configuration.RetryInterval >= 0 ? _configuration.RetryInterval : BacktraceConfiguration.DefaultRetryInterval);
```

perf2711 (commented):
Looks good. I've added one comment though.

Runtime/Model/Database/BacktraceDatabaseSettings.cs
konraddysput: I forgot about constant ! 
Let me also check (unit) cast in old .net platforms. For some reasons we decided to use convert instead.

Runtime/Model/BacktraceConfiguration.cs
```diff
@@ -28,6 +28,14 @@ public class BacktraceConfiguration : ScriptableObject
...
             UnityEngineLogLevel.Warning;
 
+        public const int DefaultRetryLimit = 3;
+        public const int DefaultReportPerMin = 50;
+        public const int DefaultGameObjectDepth = -1;
+        public const int DefaultNumberOfLogs = 10;
+        public const int DefaultMaxRecordCount = 8;
+        public const int DefaultMaxDatabaseSize = 0;
+        public const int DefaultRetryInterval = 60;
+
```
rqbacktrace, line 38: Maybe it's worth creating constants for minimum values too?

rqbacktrace (approved):
Looks good to me!

vlussenburg (approved):

perf2711 (approved):

PR #134, opened 2021-11-17
Native client compilation issues in older .NET platforms

konraddysput:
# Why

This diff fixes compilation issues on specific native platforms in older .NET versions


perf2711 (approved):
Looks good to me.

rqbacktrace (approved):
Looks good to me!

PR #135, opened 2021-11-22
Improvement/prevent duplicated anrs

konraddysput:
# Why

This diff 
1. Prevents situation when android java thread stuck and android watchdog wants to report more than one ANR
2. make sure network communication is always right - when coronerd doesn't return expected message with status code 200 (service unavailable or something) we're still able to convert JSON to data without unhandled exception,
3. Configurable ANR watchdog options

### DIFF
--- a/Android/BacktraceANRWatchdog.java
+++ b/Android/BacktraceANRWatchdog.java
@@ -62,11 +62,11 @@ public class BacktraceANRWatchdog extends Thread {
     /**
      * Initialize new instance of BacktraceANRWatchdog with default timeout
      */
-    public BacktraceANRWatchdog(String gameObjectName, String methodName) {
+    public BacktraceANRWatchdog(String gameObjectName, String methodName, int anrTimeout) {
         Log.d(LOG_TAG, "Initializing ANR watchdog");
         this.methodName = methodName;
         this.gameObjectName = gameObjectName;
-        this.timeout = DEFAULT_ANR_TIMEOUT;
+        this.timeout = anrTimeout;
         this.debug = false;
         BacktraceANRWatchdog._instance = this;
         this.start();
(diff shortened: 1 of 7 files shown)


vlussenburg (approved):

Runtime/Native/Base/NativeClientBase.cs
```diff
@@ -13,6 +13,7 @@ internal abstract class NativeClientBase
         protected const string CrashType = "Crash";
         protected const string ErrorTypeAttribute = "error.type";
 
+        protected int AnrWatchdogTimeout = 5000;
```
rqbacktrace, line 16: Should we initialize this to `DefaultAnrWatchdogTimeout`?

Android/BacktraceANRWatchdog.java
```diff
@@ -62,11 +62,11 @@
     /**
      * Initialize new instance of BacktraceANRWatchdog with default timeout
      */
-    public BacktraceANRWatchdog(String gameObjectName, String methodName) {
+    public BacktraceANRWatchdog(String gameObjectName, String methodName, int anrTimeout) {
         Log.d(LOG_TAG, "Initializing ANR watchdog");
         this.methodName = methodName;
         this.gameObjectName = gameObjectName;
-        this.timeout = DEFAULT_ANR_TIMEOUT;
```
rqbacktrace, line 69: Is this variable `DEFAULT_ANR_TIMEOUT` still used somewhere or can we delete it now? If it's still used, should we unify with `DefaultAnrWatchdogTimeout`?

Android/BacktraceANRWatchdog.java
```diff
@@ -105,6 +108,11 @@ public void run() {
                         "is on and connected debugger");
                 continue;
             }
+            if(reported) {
```
rqbacktrace: Formatting: Should it be `if (reported)`?

rqbacktrace (approved):
Looks good to me!

perf2711 (approved):
LGTM!

Android/BacktraceANRWatchdog.java
konraddysput: good point. thanks removed!

PR #139, opened 2021-12-02
Invalid background exception behavior

konraddysput:
# Why
In the past our java background exception handler prevented from passing the background thread exception information. Right now we're doing that so other handlers can handle it.

### DIFF
--- a/Android/BacktraceAndroidBackgroundUnhandledExceptionHandler.java
+++ b/Android/BacktraceAndroidBackgroundUnhandledExceptionHandler.java
@@ -16,8 +16,8 @@ public class BacktraceAndroidBackgroundUnhandledExceptionHandler implements Thre
     private final static transient String LOG_TAG = BacktraceAndroidBackgroundUnhandledExceptionHandler.class.getSimpleName();
     private final Thread.UncaughtExceptionHandler mRootHandler;
 
-    private Thread _exceptionThread;
-    private Throwable _backgroundException;
+    private Thread _lastCaughtBackgroundExceptionThread;
+    private Throwable _lastCaughtBackgroundException;
 
     /**
      * Check if data shouldn't be reported.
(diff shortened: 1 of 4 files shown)


Android/BacktraceAndroidBackgroundUnhandledExceptionHandler.java
```diff
@@ -34,15 +37,19 @@ public BacktraceAndroidBackgroundUnhandledExceptionHandler(String gameObject, St
     }
 
     @Override
-    public void uncaughtException(Thread thread, Throwable exception) {
+    public void uncaughtException(final Thread thread, final Throwable exception) {
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.CUPCAKE && mRootHandler != null && shouldStop == false) {
-            if(Looper.getMainLooper().getThread().getId() == thread.getId()) {
-                // prevent from sending exception happened to main thread - we will catch them via unity logger
-                return;
+            if (exception instanceof Exception) {
```
rqbacktrace, line 42: Why do we need this? Should we not care about Throwable Error?

Runtime/BacktraceClient.cs
```diff
@@ -991,7 +991,19 @@ internal void HandleUnhandledExceptionsFromAndroidBackgroundThread(string backgr
...
             var message = backgroundExceptionMessage.Substring(0, splitIndex);
             var stackTrace = backgroundExceptionMessage.Substring(splitIndex);
-            HandleUnityMessage(message, stackTrace, LogType.Exception);
+            if (Database != null)
+            {
+                Database.Add(new BacktraceReport(new BacktraceUnhandledException(message, stackTrace)).ToBacktraceData(null, GameObjectDepth));
+            }
+            else
+            {
+                HandleUnityMessage(message, stackTrace, LogType.Exception);
```
rqbacktrace, line 1000: Should we not always `HandleUnityMessage` for background exceptions anymore? Will this behavior be surprising for users?

rqbacktrace (approved):
Looks good to me!

Runtime/BacktraceClient.cs
vlussenburg: @rqbacktrace ; Konrad told me about this; there is a large chance it cannot be handled because Unity is dying. So prefer handling on startup via the Database.

Runtime/BacktraceClient.cs
```diff
@@ -991,7 +991,19 @@ internal void HandleUnhandledExceptionsFromAndroidBackgroundThread(string backgr
...
+            }
+            var androidNativeClient = _nativeClient as Runtime.Native.Android.NativeClient;
```
vlussenburg, line 1002: Eww. Casting. Do we need to? :) Can we fix in a different way on mainline?

vlussenburg: As in ok for bugfix, but probably want to improve leaky abstraction here

vlussenburg (commented):
Standard Vincent comment: there's no unit tests, and that makes me nervous. Can we document the testing strategy in the ticket + PR and consider adding regression test case in the example app? 

In general also please document regression tests done. Which Unity versions, which platforms, etc. Can be done via ticket and I am willing to help.

Android/BacktraceAndroidBackgroundUnhandledExceptionHandler.java
konraddysput: @rqbacktrace @vlussenburg this is a code from backtrace-android because I wanted to be sure we're doing the same thing. I think we can remove it.

Runtime/BacktraceClient.cs
konraddysput: if we have database I think it's safer to use it and send it on the next session or instead prevent application from quitting. Database storage is super fast but things might go wrong with networking AND we won't anyway apply retry settings.

Android/BacktraceAndroidBackgroundUnhandledExceptionHandler.java
```diff
@@ -16,6 +16,9 @@
     private final static transient String LOG_TAG = BacktraceAndroidBackgroundUnhandledExceptionHandler.class.getSimpleName();
     private final Thread.UncaughtExceptionHandler mRootHandler;
 
+    private Thread _exceptionThread;
```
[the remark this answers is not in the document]
konraddysput: I would say that _exceptionThread is for me cleaner than `_lastDetectedExceptionThread` or `_nonMainThreadException`

Runtime/BacktraceClient.cs
konraddysput: this is a special method that is available only for android - other native clients don't have it. This casting is safe, but I want to be sure it's always safe.

PR #140, opened 2021-12-02
Avoid using unsupported property in not-native builds

konraddysput:
# Why

Avoid using Anr Watchdog timeout in native client base on not supported platforms.

### DIFF
--- a/Runtime/Native/Base/NativeClientBase.cs
+++ b/Runtime/Native/Base/NativeClientBase.cs
@@ -1,4 +1,5 @@
-﻿using Backtrace.Unity.Model;
+﻿#if UNITY_ANDROID || UNITY_IOS || UNITY_STANDALONE_WIN
+using Backtrace.Unity.Model;
 using Backtrace.Unity.Model.Breadcrumbs;
 using Backtrace.Unity.Extensions;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


Runtime/Native/Base/NativeClientBase.cs
```diff
@@ -58,9 +58,11 @@ internal NativeClientBase(BacktraceConfiguration configuration, BacktraceBreadcr
             _configuration = configuration;
             _breadcrumbs = breadcrumbs;
             _shouldLogAnrsInBreadcrumbs = ShouldStoreAnrBreadcrumbs();
+#if UNITY_ANDROID || UNITY_IOS || UNITY_STANDALONE_WIN
```
rqbacktrace, line 61: Should all ANR code live in a new file guarded by these #ifdefs?

rqbacktrace (approved):
Looks good to me!

vlussenburg (approved):

Runtime/Native/Base/NativeClientBase.cs
konraddysput: yes, this is an option @rqbacktrace . We can put everything in the #ifdef and maybe it's a better idea because if we add a native client in the future we would need to keep in mind about anr watchdog thing.

PR #143, opened 2021-12-13
Prevent monitor disposing in the disable method

konraddysput:
# Why

Prevent a situation when the "dispose" method was called twice.

### DIFF
--- a/Runtime/Native/Android/NativeClient.cs
+++ b/Runtime/Native/Android/NativeClient.cs
@@ -163,7 +163,6 @@ namespace Backtrace.Unity.Runtime.Native.Android
             catch (Exception e)
             {
                 Debug.LogWarning(string.Format("Cannot initialize unhandled exception watcher - reason: {0}", e.Message));
-                _enabled = false;
             }
         }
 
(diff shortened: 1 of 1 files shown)


vlussenburg (approved):

Runtime/Native/Android/NativeClient.cs
```diff
@@ -504,11 +502,13 @@ public override void Disable()
             {
                 _anrWatcher.Call("stopMonitoring");
                 _anrWatcher.Dispose();
+                _anrWatcher = null;
```
perf2711, line 505: Does the method `Disable` disable this client permanently? Is there a way to re-enable it?

perf2711 (approved):
LGTM, just one thing that bugs me.

Runtime/Native/Android/NativeClient.cs
konraddysput: Right now this isn't our goal but it might be in the future. Right now this code should handle the situation when someone calls Dispose method directly via Unity API.

rqbacktrace (approved):

PR #144, opened 2021-12-13
Screen size - invalid cast

konraddysput:
# Why

`float ratio = Screen.width / Screen.height;` generates always a value `1` because of invalid type of variables (int/int => float)


vlussenburg (approved):

perf2711 (approved):
👍

rqbacktrace (approved):

PR #150, opened 2022-01-18
Android memory usage improvements

konraddysput:
# Why

This diffs improves CPU/memory usage of the backtrace-unity plugin on Android.

### DIFF
--- a/Android/BacktraceANRWatchdog.java
+++ b/Android/BacktraceANRWatchdog.java
@@ -63,11 +63,14 @@ public class BacktraceANRWatchdog extends Thread {
      */
     @Override
     public void run() {
-        Boolean reported = false;
-        Log.d(LOG_TAG, "Starting ANR watchdog. Anr timeout: " + this.timeout);
         if (Debug.isDebuggerConnected() || Debug.waitingForDebugger()) {
+            Log.d(LOG_TAG, "Detected a debugger connection. ANR Watchdog is disabled");
             return;
         }
+        
+        Boolean reported = false;
+        Log.d(LOG_TAG, "Starting ANR watchdog. Anr timeout: " + this.timeout);
+
         while (!shouldStop && !isInterrupted()) {
             final backtrace.io.backtrace_unity_android_plugin.BacktraceThreadWatcher threadWatcher = new backtrace.io.backtrace_unity_android_plugin.BacktraceThreadWatcher(0, 0);
             mainThreadHandler.post(new Runnable() {
--- a/Android/BacktraceANRWatchdog.java
+++ b/Android/BacktraceANRWatchdog.java
@@ -17,15 +17,8 @@ import java.util.Calendar;
  */
 public class BacktraceANRWatchdog extends Thread {
 
-    private static BacktraceANRWatchdog _instance;
-
     private final static transient String LOG_TAG = BacktraceANRWatchdog.class.getSimpleName();
 
-
-    /**
-     * Enable debug mode - errors will not be sent if the debugger is connected
-     */
-    private final boolean debug;
     /**
      * Handler for UI Thread - used to check if the thread is not blocked
      */
@@ -62,8 +55,6 @@ public class BacktraceANRWatchdog extends Thread {
         this.methodName = methodName;
         this.gameObjectName = gameObjectName;
         this.timeout = anrTimeout;
-        this.debug = false;
-        BacktraceANRWatchdog._instance = this;
         this.start();
     }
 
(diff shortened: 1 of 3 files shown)


Android/BacktraceANRWatchdog.java
```diff
@@ -74,8 +65,10 @@ public BacktraceANRWatchdog(String gameObjectName, String methodName, int anrTim
     public void run() {
         Boolean reported = false;
         Log.d(LOG_TAG, "Starting ANR watchdog. Anr timeout: " + this.timeout);
+        if (Debug.isDebuggerConnected() || Debug.waitingForDebugger()) {
+            return;
```
rqbacktrace, line 69: Maybe we should keep a log message here?

Android/BacktraceANRWatchdog.java
```diff
@@ -74,8 +65,10 @@ public BacktraceANRWatchdog(String gameObjectName, String methodName, int anrTim
...
+        }
         while (!shouldStop && !isInterrupted()) {
```
rqbacktrace, line 71: We should put a newline before this

Runtime/Native/Android/NativeClient.cs
```diff
@@ -431,7 +431,9 @@ public void HandleAnr()
 
                                     NativeReport(AndroidJNI.NewStringUTF(AnrMessage), true);
                                     // update error.type attribute in case when crash happen 
-                                    SetAttribute(ErrorTypeAttribute, CrashType);
+                                    AddAttribute(
```
rqbacktrace, line 434: Isn't it more safe to use `SetAttribute` here? We can do the same for `HangType` attribute as well.

It shouldn't increase baseline RAM usage either since we only take this code path when we create an ANR report.

konraddysput: The AddAttribute method gonna do some additional checks, that are unnecessary in this situation because we're sure that attribute name and attribute value is always defined.

vlussenburg:
Could be get an indication on what difference this makes on a few devices? Might I suggest adding the following data to perhaps the Performance Metrics (and maybe add to Breadcrumbs?).

```
    Debug.Log("[PERFORMANCE][BEFORE] Total Reserved memory by Unity: " + Profiler.GetTotalReservedMemoryLong() + "Bytes");
    Debug.Log("[PERFORMANCE][BEFORE] Allocated memory by Unity: " + Profiler.GetTotalAllocatedMemoryLong() + "Bytes");
    Debug.Log("[PERFORMANCE][BEFORE] Reserved but not allocated: " + Profiler.GetTotalUnusedReservedMemoryLong() + "Bytes");
    Debug.Log("[PERFORMANCE][BEFORE] HEAP : " + System.GC.GetTotalMemory(false)  + "Bytes"); 

    _backtraceClient = BacktraceClient.Initialize(
        backTraceConfig,
        attributes);
    Profiler.EndSample();

    Debug.Log("[PERFORMANCE] ELAPSED START" + sp.ElapsedMilliseconds);
    Debug.Log("[PERFORMANCE][AFTER] Total Reserved memory by Unity: " + Profiler.GetTotalReservedMemoryLong() + "Bytes");
    Debug.Log("[PERFORMANCE][AFTER] Allocated memory by Unity: " + Profiler.GetTotalAllocatedMemoryLong() + "Bytes");
    Debug.Log("[PERFORMANCE][AFTER] Reserved but not allocated: " + Profiler.GetTotalUnusedReservedMemoryLong() + "Bytes");
    Debug.Log("[PERFORMANCE][AFTER] HEAP : " + System.GC.GetTotalMemory(false)  + "Bytes"); 
```

konraddysput:
> Could be get an indication on what difference this makes on a few devices? Might I suggest adding the following data to perhaps the Performance Metrics (and maybe add to Breadcrumbs?).
> 
> (I know you can't use the exact code, but this is an example on how another game tested our memory/time impact)
> 
> ```
>     Debug.Log("[PERFORMANCE][BEFORE] Total Reserved memory by Unity: " + Profiler.GetTotalReservedMemoryLong() + "Bytes");
>     Debug.Log("[PERFORMANCE][BEFORE] Allocated memory by Unity: " + Profiler.GetTotalAllocatedMemoryLong() + "Bytes");
>     Debug.Log("[PERFORMANCE][BEFORE] Reserved but not allocated: " + Profiler.GetTotalUnusedReservedMemoryLong() + "Bytes");
>     Debug.Log("[PERFORMANCE][BEFORE] HEAP : " + System.GC.GetTotalMemory(false)  + "Bytes"); 
> 
>     _backtraceClient = BacktraceClient.Initialize(
>         backTraceConfig,
>         attributes);
>     Profiler.EndSample();
> 
>     Debug.Log("[PERFORMANCE] ELAPSED START" + sp.ElapsedMilliseconds);
>     Debug.Log("[PERFORMANCE][AFTER] Total Reserved memory by Unity: " + Profiler.GetTotalReservedMemoryLong() + "Bytes");
>     Debug.Log("[PERFORMANCE][AFTER] Allocated memory by Unity: " + Profiler.GetTotalAllocatedMemoryLong() + "Bytes");
>     Debug.Log("[PERFORMANCE][AFTER] Reserved but not allocated: " + Profiler.GetTotalUnusedReservedMemoryLong() + "Bytes");
>     Debug.Log("[PERFORMANCE][AFTER] HEAP : " + System.GC.GetTotalMemory(false)  + "Bytes"); 
> ```

I think we can use them for testing purposes, but we should use them in our backtrace-unity code. This code belongs to the test project instead.

Android/BacktraceANRWatchdog.java
konraddysput: done, thx.

PR #151, opened 2022-01-19
fix unable to delete reports when db path has trailing slash

sarahedkins:


konraddysput (approved):
Looks good to me. Thanks!

PR #155, opened 2022-02-16
Game initialization order improvements

konraddysput:
# Why

This diff runs game object initialization in the right order with all rules that we already set for Initialize/GameObject API.


Runtime/BacktraceDatabase.cs
```diff
@@ -182,7 +192,7 @@ public void Reload()
 #if UNITY_SWITCH
```
perf2711, line 192: small nit: Maybe check the Switch condition before validating the configuration?

Runtime/BacktraceDatabase.cs
```diff
@@ -622,7 +754,21 @@ protected virtual void LoadReports()
             {
                 return;
             }
+
             var files = BacktraceDatabaseFileContext.GetRecords();
+            if (!files.Any())
```
perf2711, line 759: `GetRecords` returns `IEnumerable`. You enumerate the collection twice, here, and in `foreach`. Maybe materialize the collection by calling `.ToArray()`?

Runtime/Model/Breadcrumbs/BacktraceBreadcrumbs.cs
```diff
@@ -191,5 +196,14 @@ public static bool CanStoreBreadcrumbs(UnityEngineLogLevel logLevel, BacktraceBr
         {
             return backtraceBreadcrumbsLevel != BacktraceBreadcrumbType.None && logLevel != UnityEngineLogLevel.None;
         }
+
+        public string Archive()
+        {
+            if (string.IsNullOrEmpty(_archivePath))
```
perf2711, line 202: This will archive only once in the lifetime of the class. Is this intended?

Runtime/Model/Breadcrumbs/InMemory/BacktraceInMemoryLogManager.cs
```diff
@@ -46,6 +46,15 @@ public string BreadcrumbsFilePath
             }
         }
 
+        /// <summary>
+        /// Archive in memory breadcrumbs. For in-memory breadcrumbs this operation is unsupported.
+        /// </summary>
+        /// <returns>Empty string</returns>
+        public string Archive()
```
perf2711, line 53: If you want to make unsupported features, why not introduce a new interface, like `IArchivable`? That way, you don't have to make empty methods that return empty strings, and when you want to use the `Archive` method, you would have to check if the class supports it (`if (instance is IArchivable)`).

Runtime/Native/INativeClient.cs
```diff
@@ -41,5 +41,10 @@ internal interface INativeClient : IDynamicAttributeProvider
         /// <param name="state">True if should pause, otherwise false.</param>
         void PauseAnrThread(bool state);
 
+        /// <summary>
+        /// Determine if the native integration layer is initialized
+        /// </summary>
+        /// <returns>True, if the native client is enabled</returns>
+        bool Initialized();
```
perf2711, line 48: small nit: shouldn't this be a getter?

```
bool Initialized { get; }
```

If not, maybe name the method `IsInitialized`?

perf2711 (approved):
LGTM. Added some comments.

Runtime/BacktraceClient.cs
```diff
@@ -525,52 +562,16 @@ public void Refresh()
...
-                // breadcrumb path is required by native integration and will be added just before native integration initialization
-                if (!string.IsNullOrEmpty(breadcrumbsPath))
+                _database = GetComponent<BacktraceDatabase>();
+                if (_database != null)
                 {
-                    nativeAttachments.Add(breadcrumbsPath);
+                    _database.SetApi(BacktraceApi);
+                    _database.SetReportWatcher(_reportLimitWatcher);
+                    _database.Reload();
+                    _breadcrumbs = Database.Breadcrumbs as BacktraceBreadcrumbs;
```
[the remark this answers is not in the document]
konraddysput: Yes, I expect that this cast might return null - it's expected to safety cast it to null.

konraddysput:
We divided this pull request into few smaller ones.

PR #157, opened 2022-05-04
Adds a checkbox for iOS builds to enable client side unwinding.

perf2711:
A simple change that adds the "Enable client side unwinding" checkbox, available on Android, to iOS builds.

At this moment, the option is passed to the `NativeClient`'s `Start` method, but we need to wait for a native implementation of this feature.

### DIFF
--- a/Runtime/Native/iOS/NativeClient.cs
+++ b/Runtime/Native/iOS/NativeClient.cs
@@ -63,7 +63,7 @@ namespace Backtrace.Unity.Runtime.Native.iOS
             }
             if (_configuration.CaptureNativeCrashes)
             {
-                HandleNativeCrashes(clientAttributes, attachments, configuration.ClientSideUnwinding);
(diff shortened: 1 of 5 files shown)


Runtime/Native/iOS/NativeClient.cs
```diff
@@ -63,7 +63,7 @@ public NativeClient(BacktraceConfiguration configuration, BacktraceBreadcrumbs b
             }
             if (_configuration.CaptureNativeCrashes)
             {
-                HandleNativeCrashes(clientAttributes, attachments);
+                HandleNativeCrashes(clientAttributes, attachments, configuration.ClientSideUnwinding);
```
konraddysput, line 66: I think we don't need to pass it there via parameter - we have an access to the _configuration object anyway in the method body.

konraddysput (approved):
Looks good - let's also cover the version bump and release notes. I think we need to talk about this with @vlussenburg

PR #159, opened 2022-07-04
Bugfix/next session breadcrumbs path

konraddysput:
# Why

We discovered a problem with breadcrumb support when we finished the previous session. If the game can't send errors on the next session startup because the game is delayed because of the startup jobs, then we probably won't have a breadcrumb file associated with the report. This diff makes a copy of the existing breadcrumb file on the session startup and replaces the old path to the breadcrumb file with a new one. In addition, this diff also delays possible breadcrumbs that backtrace-unity will store. Because of that, we're sure that we won't override the breadcrumb file before making a copy.

This diff is really similar to the diff that I already submitted while ago. In this diff I only focus on the breadcrumb problem and this diff doesn't touch initialization order - so it will be much safer to land if earlier.

### DIFF
--- a/Runtime/Model/Breadcrumbs/Storage/BacktraceStorageLogManager.cs
+++ b/Runtime/Model/Breadcrumbs/Storage/BacktraceStorageLogManager.cs
@@ -331,7 +331,8 @@ namespace Backtrace.Unity.Model.Breadcrumbs.Storage
                 return string.Empty;
             }
 
-            var copyPath = Path.Combine(_storagePath, string.Format("{0}-1", BreadcrumbLogFilePrefix));
+            const string archivePattern = "{0}-1";
+            var copyPath = Path.Combine(_storagePath, string.Format(archivePattern, BreadcrumbLogFilePrefix));
             if (File.Exists(copyPath))
             {
                 File.Delete(copyPath);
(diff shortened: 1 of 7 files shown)


vlussenburg (approved):

Runtime/Model/Breadcrumbs/Storage/BacktraceStorageLogManager.cs
```diff
@@ -317,5 +323,22 @@ public double BreadcrumbId()
...
+                return string.Empty;
+            }
+
+            const string archivePattern = "{0}-1";
+            var copyPath = Path.Combine(_storagePath, string.Format(archivePattern, BreadcrumbLogFilePrefix));
+            if (File.Exists(copyPath))
+            {
+                File.Delete(copyPath);
+            }
+            File.Copy(BreadcrumbsFilePath, copyPath);
```
perf2711, line 340: I think `File.Copy` has an overload that supports overwriting the file:

https://docs.microsoft.com/pl-pl/dotnet/api/system.io.file.copy?view=netframework-2.0#system-io-file-copy(system-string-system-string-system-boolean)

It's supported even in .NET Framework 2.0, so you should be able to use it here.

perf2711 (approved):
LGTM, only small nit.

Runtime/Model/Breadcrumbs/Storage/BacktraceStorageLogManager.cs
konraddysput: fixed. thx!

PR #161, opened 2022-07-08
Breadcrumbs support for windows after the app restart/crash

konraddysput:
# Why

On the previous diff, we added a multi-session breadcrumbs support. In the past, we added also a similar solution for windows native dumps captured by the unity crash handler. In this diff, I tried to fix the flow for windows native and also simplify the code/remove redundant code. 

# testing strategy
1.  Crash app on Windows that will be captured by unity crash handler. See breadcrumbs on UI
2. Crash app on WIndows that will be captured by Backtrace. See breadcrumbs on UI
3. Crash app on Android - see breadcrumbs on UI


Runtime/BacktraceClient.cs
```diff
@@ -1318,5 +1314,13 @@ private bool ShouldSkipReport(ReportFilterType type, Exception exception, string
                 || (SkipReport != null && SkipReport.Invoke(type, exception, message));
 
         }
+
+        internal IList<string> GetNativeAttachments()
+        {
+            return _clientReportAttachments.ToList()
```
perf2711, line 1320: is the first `.ToList()` here necessary?

Runtime/BacktraceDatabase.cs
```diff
@@ -261,10 +264,39 @@ private void Start()
...
+            var isEnabled = Enable && Configuration.SendUnhandledGameCrashesOnGameStartup && isActiveAndEnabled;
+            var hasNativeConfiguration = _client && _client.NativeClient != null && _client.NativeClient is IStartupMinidumpSender;
+            if (isEnabled && hasNativeConfiguration)
+            {
+                var client = _client.NativeClient as IStartupMinidumpSender;
+                var attachments = _client.GetNativeAttachments();
+                if(string.IsNullOrEmpty(breadcrumbArchive))
+                {
+                    attachments.Add(breadcrumbArchive);
+                }
```
perf2711, line 294: is this valid? Do you want to add `breadcrumbArchive` only if it is `null` or empty?

vlussenburg (approved):

Runtime/BacktraceClient.cs
konraddysput: you're right - we don't need it. Thank you.

Runtime/BacktraceDatabase.cs
konraddysput: I messed this up before finalizing pull request. Now it should work. Thank you!

perf2711 (approved):

PR #162, opened 2022-07-08
Generate not-empty machine identifier

konraddysput:
# Why

Recently on some consoles, we recognized empty machine identifiers. In this diff I added a new machine storage id class that makes sure that we always have:
* not empty guid generated based on the Unity/Networking methods
* not empty guid on platforms that not support Unity/Networking interfaces
* uuid storage over the session - so for platforms where the uuid needs to be generated, the uuid will stay the same.


Tests/Runtime/Client/BacktraceAttributeMachineIdTests.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        [Test]
+        public void TestMachineAttributes_ShouldUseMac_ShouldReturnNetowrkingIdentifier()
+        {
+            var machineIdStorage = new MachineIdStorageMock(false);
+
+            var machineId = machineIdStorage.GenerateMachineId();
+
+            Assert.IsFalse(GuidHelper.IsEmptyGuid(machineId));
+        }
```
perf2711, line 35: Maybe in these two, you should test if the value does not change after clearing PlayerPrefs key?

Runtime/Common/GuidHelper.cs
```diff
@@ -17,5 +17,11 @@ public static Guid FromLong(long source)
             Array.Copy(BitConverter.GetBytes(source), guidData, 8);
             return new Guid(guidData);
         }
+
+        public static bool IsEmptyGuid(string guid)
+        {
+            const string emptyGuid = "00000000-0000-0000-0000-000000000000";
+            return string.IsNullOrEmpty(guid) || guid == emptyGuid;
```
vlussenburg, line 24: Should you use .equals for comparison in c#?
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/equality-operators#string-equality

Ok, the answer is no, never mind :)

Runtime/Model/MachineIdStorage.cs
```diff
@@ -0,0 +1,107 @@
...
+            {
+                return null;
+            }
+            return SystemInfo.deviceUniqueIdentifier;
+        }
+
+        /// <summary>
+        /// Use Networking interface to generate machine identifier - MAC number from the networking interface.
+        /// </summary>
+        /// <returns>Machine id - MAC in a GUID format. If the networking interface is not available then it returns empty string.</returns>
```
vlussenburg, line 87: It returns null, not empty string

Runtime/Model/MachineIdStorage.cs
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        /// <summary>
+        /// Use Unity device identifier to generate machine identifier
+        /// </summary>
+        /// <returns>Unity machine identifier if the device identifier is supported. Otherwise null</returns>
+        protected virtual string UseUnityIdentifier()
+        {
+            // DeviceUniqueIdentifier will return "Switch" on Nintendo Switch
+            // try to generate random guid instead
```
vlussenburg, line 76: This comment doesn't seem to match what the code does

vlussenburg: Also, I think through the ifdef this code is never reached for Switch

Runtime/Model/MachineIdStorage.cs
```diff
@@ -0,0 +1,107 @@
...
+            var storageMachineId = FetchMachineIdFromStorage();
+            if (!string.IsNullOrEmpty(storageMachineId))
+            {
+                return storageMachineId;
+            }
+#if !UNITY_WEBGL && !UNITY_SWITCH
+            var unityIdentifier = UseUnityIdentifier();
+            if (!GuidHelper.IsEmptyGuid(unityIdentifier))
+            {
+                StoreMachineId(unityIdentifier);
```
vlussenburg, line 35: Programming flow nitpicky-ness: `StoreMachineId` is called in multiple locations. Why not assign one `machineId` var and store it in the end instead of doing an early out? Downside is that you'll have a large if/elseif/elseif/ellse/ tree, so, you can probably ignore this.

Runtime/Common/GuidHelper.cs
```diff
@@ -17,5 +17,11 @@ public static Guid FromLong(long source)
             Array.Copy(BitConverter.GetBytes(source), guidData, 8);
             return new Guid(guidData);
         }
+
+        public static bool IsEmptyGuid(string guid)
```
vlussenburg, line 21: Method should be `isEmptyOrNullOrEmptyGuid`.

Tests/Runtime/Client/BacktraceAttributeMachineIdTests.cs
```diff
@@ -0,0 +1,74 @@
+﻿using Backtrace.Unity.Extensions;
+using Backtrace.Unity.Model;
+using Backtrace.Unity.Tests.Runtime.Client.Mocks;
+using NUnit.Framework;
+using UnityEngine;
+
+namespace Backtrace.Unity.Tests.Runtime.Client
+{
+    class BacktraceAttributeMachineIdTests
```
vlussenburg, line 9: Nice tests! <3

vlussenburg (approved):

Runtime/Common/GuidHelper.cs
konraddysput: `== ` == `Equals

Runtime/Model/MachineIdStorage.cs
konraddysput: done. thx

Runtime/Model/MachineIdStorage.cs
konraddysput: the flow makes sure this code won't be run against WebGL and Switch

Runtime/Model/MachineIdStorage.cs
konraddysput: the logic would be harder to understand with more ifs to check. Now it's kinda clear.

Runtime/Common/GuidHelper.cs
konraddysput: The class name is GuidHelper - the domain of the method is clear, but your suggestion anyway is really good. I'm going to rename this method to stick to .NET patterns and use something like "IsNullOrEmpty"

perf2711 (approved):

PR #164, opened 2022-08-02
Package attribute

konraddysput:
# Why

This diff adds `application.package` attribute to list of default backtrace-unity attributes. The attribute will be now available on every type of report


Runtime/Model/Attributes/RuntimeAttributeProvider.cs
```diff
@@ -22,6 +22,7 @@ public void GetAttributes(IDictionary<string, string> attributes)
             attributes["application.company.name"] = Application.companyName;
             attributes["application.data_path"] = Application.dataPath;
             attributes["application.id"] = Application.identifier;
+            attributes["application.package"] = Application.identifier;
```
perf2711, line 25: is this supposed to have the same value as `"application.id"`?

konraddysput: yeah

perf2711 (approved):

PR #172, opened 2022-10-11
Fix inalid stack frame type

konraddysput:
# Why

By default, if the stack trace algorithm cannot find a correct way to determine the OS, the algorithm uses "Android" parser to parse the stack frame. This diff changes the behavior of the stack trace parser. Now we're guessing if we can use "android" parser and depends on the if def condition


vlussenburg (approved):
LGTM

PR #174, opened 2022-11-04
Replace crashpad handler

konraddysput:
# Why

This diff rollbacks the armeabi crashpad handler to the previous version


vlussenburg (changes requested):
Acceptable as workaround

PR #187, opened 2023-08-29
Bug fix and performance improvement

SGSmithI:
- Bug: FlushRecord not starting stopwatch.
- Removed unecessary allocation of stopwatch when unused.

Edit (removed point):
- Removed excessive allocation of enumerators and loops over attachments enumerable in Add

### DIFF
--- a/Runtime/BacktraceDatabase.cs
+++ b/Runtime/BacktraceDatabase.cs
@@ -381,11 +381,11 @@ namespace Backtrace.Unity
 
             //add built-in attachments
             var attachments = BacktraceDatabaseFileContext.GenerateRecordAttachments(data);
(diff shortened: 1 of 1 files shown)


konraddysput:
Are you sure you included all your code?

konraddysput:
The reason why we have this code is to avoid the potential allocation of any element via foreach - if you inspect the code in the profiles, each foreach invocation allocates an object. To make sure we work as fast as possible and we do not generate any unnecessary allocations, the library uses for instead of foreach.

From the developer perspective, it's better to use foreach. In the critical library paths executed every time you generate a report, for was better. 

Based on your code I don't see this:
`Removed excessive allocation of enumerators and loops over attachments enumerable in Add`

Can you please elaborate more?

PR #189, opened 2023-10-27
Use C# api to upload symbols

konraddysput:
This diff fixes the symbols archive - because of recent changes the command line tool we used to archive/modify symbols is not available. Now we're using pure C# API to do such a thing.

### DIFF
--- a/Editor/Native/Android/SymbolsUpload.cs
+++ b/Editor/Native/Android/SymbolsUpload.cs
@@ -93,7 +93,21 @@ namespace Backtrace.Unity.Editor.Build
 
             try
             {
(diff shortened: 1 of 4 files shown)


rick-bt (approved):

Editor/Native/Android/SymbolsUpload.cs
```diff
@@ -93,36 +93,15 @@ private string ConvertSymbols(string symbolsArchive)
 
             try
             {
-                var unpackProcess = new System.Diagnostics.Process()
```
konraddysput, line 96: we still need to use it if we want to support older Unity envs

PR #194, opened 2023-11-08
metrics: retry submission on connection failure

gm4sl:
Previously, only a subset protocol errors would be considered for retry when submitting metrics (summed or unique) to events.backtrace.io. This excluded, notably, a failure to connect to the server at all. If the device was offline at startup, for instance, then the "Application Launches" event would never get sent.

This commit changes the handling so that any network error will also go through the retry logic for submitting an event.

### DIFF
--- a/Runtime/Model/Metrics/MetricsSubmissionQueue.cs
+++ b/Runtime/Model/Metrics/MetricsSubmissionQueue.cs
@@ -116,8 +116,11 @@ namespace Backtrace.Unity.Model.Metrics
                 {
                     OnRequestCompleted();
                 }
-                else if (statusCode > 501 && statusCode != 505)
+                else if (httpError == true || (statusCode > 501 && statusCode != 505))
                 {
+                    // Failed to communicate with server or received a 5xx response code. Retry
+                    // again at a later time, up to the configured number of maximum attempts.
+                    //
                     _numberOfDroppedRequests++;
                     if (attempts + 1 == BacktraceMetrics.MaxNumberOfAttempts)
                     {


Runtime/Model/Metrics/MetricsSubmissionQueue.cs
```diff
@@ -112,12 +112,15 @@ internal void SendPayload(ICollection<T> events, uint attempts = 0)
 
             RequestHandler.Post(SubmissionUrl, payload, (long statusCode, bool httpError, string response) =>
             {
-                if (statusCode == 200)
+                if (httpError == false && statusCode == 200)
                 {
                     OnRequestCompleted();
                 }
-                else if (statusCode > 501 && statusCode != 505)
+                else
```
konraddysput, line 119: I'm not sure about that. Those statuses were confirmed with the team and Paul when we were working on the integration. In the situation when status code = 403/404 it doesn't make sense to retry.

Runtime/Model/Metrics/MetricsSubmissionQueue.cs
```diff
@@ -112,12 +112,15 @@ internal void SendPayload(ICollection<T> events, uint attempts = 0)
 
             RequestHandler.Post(SubmissionUrl, payload, (long statusCode, bool httpError, string response) =>
             {
-                if (statusCode == 200)
+                if (httpError == false && statusCode == 200)
```
konraddysput, line 115: snippet on how we generate each parameter: 

```
var statusCode = request.responseCode;
                var response = request.downloadHandler.text;
                var networkError = request.ReceivedNetworkError();
```

if the status code == 200 there is no way the error will be there. EVen if the error is set, 200 means the server received and accepted the request, so there is no reason to retry it.

konraddysput (changes requested):
I strongly believe we don't need those changes. Please see my comments

Runtime/Model/Metrics/MetricsSubmissionQueue.cs
gm4sl: Ok. What about when `httpError == true`? In this case, the metric is never sent. This can happen if the connection is temporarily down for instance.

Runtime/Model/Metrics/MetricsSubmissionQueue.cs
gm4sl: Agreed. Thanks.

Runtime/Model/Metrics/MetricsSubmissionQueue.cs
konraddysput: how about extending the check by checking if `httpError == true || (statusCode > 501 && statusCode != 505)` ?

Runtime/Model/Metrics/MetricsSubmissionQueue.cs
```diff
@@ -116,8 +116,11 @@ internal void SendPayload(ICollection<T> events, uint attempts = 0)
                 {
                     OnRequestCompleted();
                 }
-                else if (statusCode > 501 && statusCode != 505)
+                else if (httpError == true || (statusCode > 501 && statusCode != 505))
                 {
+                    // Failed to communicate with server or received 4xx/5xx response code. Retry
```
konraddysput, line 121: I think we also need to adjust the comment - we do not care about 4xx responses anymore, but we do care about networking issues or connection resets (or HTTP connection errors in general)

konraddysput (changes requested):

Runtime/Model/Metrics/MetricsSubmissionQueue.cs
gm4sl: Updated. Thank you.

konraddysput (approved):

PR #197, opened 2023-11-15
Added the screencapture module as a dependency

NordlightPer:
The package actually needs the screen capture module to be present and while that is included in manifest.json by default, it can be removed

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,5 +20,8 @@
     "url": "https://backtrace.io/"
   },
   "category": "Editor plugin + Script plugin",
-  "type": "library"
+  "type": "library",
+  "dependencies": {
+    "com.unity.modules.screencapture": "1.0.0"
+  }
 }


konraddysput:
hey @NordlightPer 

I cannot find anything about this change in the Unity documentation and when we were implementing this feature, the library was available by default in the editor - do you have any resources i can read to confirm we need to start using the package?

NordlightPer:
> hey @NordlightPer
> 
> I cannot find anything about this change in the Unity documentation and when we were implementing this feature, the library was available by default in the editor - do you have any resources i can read to confirm we need to start using the package?

When you create a new Unity project you get an autogenerated manifest.json file that (just tested in 2022.3) contains about 40 internal Unity packages. One of them is `"com.unity.modules.screencapture": "1.0.0"` 
Since I don't think it makes sense to include a bunch of modules we don't use, I have cleaned that list up and when integrating Backtrace-unity I got compilation errors since you are using the screen capture. 
My PR doesn't really do anything but stating this dependency in the package.json so that it will be automatically resolved as a transitive dependency when adding backtrace-unity.

konraddysput (approved):

PR #198, opened 2023-12-13
Xbox native client improvements

konraddysput:
# Why

In the xbox native client implementation, we have found a few places that require adjustments. This code makes sure:
- the error.type attribute is set like in every other native client implementation
- attributes are set only when the client is enabled,
- the integration is starting when the database directory exists.


gm4sl (approved):
Looks good. Verified changes on Xbox devkit.

PR #216, opened 2024-06-12
Proguard support

konraddysput:
# Why

This pull request adds support for Proguard to the `BacktraceClient`. By using our new proguard symbolication API our customers can set symbolication id for their proguard maps.

ref: BT-2975

### DIFF
--- a/Runtime/BacktraceClient.cs
+++ b/Runtime/BacktraceClient.cs
@@ -66,6 +66,15 @@ namespace Backtrace.Unity
             }
         }
 
+#if UNITY_ANDROID
+        private bool _useProguard = false;
+
+        public void UseProguard(String symbolicationId) {
+            _useProguard = true;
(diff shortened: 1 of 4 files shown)


Runtime/BacktraceClient.cs
```diff
@@ -988,15 +1001,20 @@ internal void HandleUnhandledExceptionsFromAndroidBackgroundThread(string backgr
             }
             var message = backgroundExceptionMessage.Substring(0, splitIndex);
             var stackTrace = backgroundExceptionMessage.Substring(splitIndex);
+            var report = new BacktraceReport(new BacktraceUnhandledException(message, stackTrace));
+             if (_useProguard) {
```
perf2711, line 1005: nit: formatting

perf2711 (approved):

PR #218, opened 2024-06-14
Use Java crash handler on not extracted libraries

konraddysput:
# Why

This update allows using java crash handler when the native libraries are not extracted. This change only works on Android. 

In the current implementation, we're using Java crash handler as a fallback function. In most cases, we know that our native crash handler should be available and we know most of our clients might not have proguard configuration set up by default


Runtime/Model/Attributes/MachineAttributeProvider.cs
```diff
@@ -58,12 +58,23 @@ private void IncludeOsInformation(IDictionary<string, string> attributes)
...
-                attributes["uname.version"] = version.GetStatic<string>("RELEASE").ToString();
+                    var deviceSdkVersion = version.GetStatic<int>("SDK_INT");
+                    attributes["device.sdk"] = deviceSdkVersion.ToString();
+                    if(deviceSdkVersion >= 21) 
+                    {
+                        string[] supportedAbis = build.GetStatic<string[]>("SUPPORTED_ABIS");
+                        
+                        if (supportedAbis != null && supportedAbis.Length > 0)
+                        {
+                            attributes["device.abi"] =  supportedAbis[0];
```
rick-bt, line 73: Does this need ToString?
```suggestion
                            attributes["device.abi"] =  supportedAbis[0].ToString();

```

konraddysput: No, it's already an array of strings.

rick-bt (commented):
Runtime/Native/Android/NativeClient.cs#255
When we `Directory.CreateDirectory(databasePath)`, can we check that it was created successfully? Would be helpful to know if that create failed.

rick-bt (approved):

konraddysput:
> Runtime/Native/Android/NativeClient.cs#255 When we `Directory.CreateDirectory(databasePath)`, can we check that it was created successfully? Would be helpful to know if that create failed.

Directory.CreateDirectory ensures that the directory exists after it is called. Its in the c# implementation

Android/BacktraceCrashHandler.java
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        String crashHandlerLibrary = environmentVariables.get(BACKTRACE_CRASH_HANDLER);
+        if (crashHandlerLibrary == null) {
+            Log.e(LOG_TAG, String.format("Cannot capture crash dump. Cannot find %s environment variable", BACKTRACE_CRASH_HANDLER));
+            return false;
+        }
+        System.load(crashHandlerLibrary);
+
+        boolean result = handleCrash(args);
```
perf2711, line 32: does it make sense to try/catch this?

Runtime/Native/Android/NativeClient.cs
```diff
@@ -309,6 +310,71 @@ private void HandleNativeCrashes(IDictionary<string, string> backtraceAttributes
...
+            var envVariableDictionary =  Environment.GetEnvironmentVariables();
+            if (envVariableDictionary == null) {
+                Debug.LogWarning("Environment variables are not defined.");
+                return false;
+            }
+
+            // verify if the library is already extracted
+            var backtraceNativeLibraryPath = Path.Combine(nativeDirectory, _nativeLibraryName);
+            if (!File.Exists(backtraceNativeLibraryPath)) {
+                backtraceNativeLibraryPath = string.Format("{0}!/lib/{1}/{2}", Application.dataPath, deviceAbi, _nativeLibraryName);
```
perf2711, line 344: why not use `Path.Combine` here? Also, is the `!` required here?

Android/BacktraceCrashHandler.java
konraddysput: this code shouldn't crash. in addition to that, we cannot do anything with an exception, unless you have an idea what to do. 
This
 code invokes crashpad, and our native interface provides a boolean response. If anything happens in the NDK layer, most likely it will crash the app, so try/catch won't handle it.

Also - since this is a class invoked by the external command, it would be better if we provide a full description of the exception with parameters. Wrapping it into a try/catch block with some nice messages won't help.

Runtime/Native/Android/NativeClient.cs
konraddysput: this is not a typical path - since `!` points to the path inside the apk. Thats why I wanted to use Format instead.

perf2711 (approved):

PR #225, opened 2024-08-30
Feature/mac support

melekr:
- Adds native macOS support

ref: BT-1029


Runtime/Model/Attributes/MachineAttributeProvider.cs
```diff
@@ -90,6 +90,10 @@ private void IncludeOsInformation(IDictionary<string, string> attributes)
             attributes["uname.version"] = UnityEngine.iOS.Device.systemVersion;
             attributes["uname.fullname"] = Environment.OSVersion.Version.ToString();
 #endif
+
```
konraddysput, line 93: why we cannot use the if statement above ?

Runtime/Native/NativeClientFactory.cs
```diff
@@ -22,8 +22,10 @@ internal static INativeClient CreateNativeClient(BacktraceConfiguration configur
             return new Android.NativeClient(configuration, breadcrumbs, attributes, attachments, gameObjectName);
 #elif UNITY_IOS
             return new iOS.NativeClient(configuration, breadcrumbs, attributes, attachments);
+#elif UNITY_STANDALONE_OSX
+            return new OSX.NativeClient(configuration, breadcrumbs, attributes, attachments);
 #else
-            return null;
+                return null;
```
konraddysput, line 28: additional tab

konraddysput (commented):
Did you verify OOMs?

Runtime/Model/Attributes/MachineAttributeProvider.cs
melekr: UnityEngine.iOS.Device.systemVersion is iOS specific

melekr:
> Did you verify OOMs?

OOM generates a crash report with signal code : SEGV_ACCERR and Error Message: Segmentation fault: 11.

Runtime/Model/Attributes/MachineAttributeProvider.cs
konraddysput: Yeah but in talking about environment.osversion

konraddysput (approved):

PR #232, opened 2025-01-10
Expose BacktraceMetrics.SessionId and BacktraceClient.AttributeProvider

hymerman:
This is useful to allow SessionId and other attributes (e.g. 'guid') to be written to 'dying memory' on platforms that support it (e.g. Switch, PlayStation), which the Backtrace server is then able to parse on importing from platform crash reporting services, and set on the imported errors. It can then be correlated with other errors from the same session.

### DIFF
--- a/Runtime/BacktraceClient.cs
+++ b/Runtime/BacktraceClient.cs
@@ -50,7 +50,7 @@ namespace Backtrace.Unity
         /// <summary>
         /// Client attribute provider
         /// </summary>
-        internal AttributeProvider AttributeProvider
+        public AttributeProvider AttributeProvider
         {
             get
             {
--- a/Runtime/Interfaces/IBacktraceMetrics.cs
+++ b/Runtime/Interfaces/IBacktraceMetrics.cs
@@ -1,4 +1,5 @@
-﻿using Backtrace.Unity.Model.Attributes;
+﻿using System;
+using Backtrace.Unity.Model.Attributes;
 using Backtrace.Unity.Model.Metrics;
 using System.Collections.Generic;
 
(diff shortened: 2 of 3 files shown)


konraddysput:
hey @hymerman ! 
Thanks for contributing. Why can you not get all the attributes instead and in case you're interested only in some of them, select the most interesting for you?

hymerman:
Thanks!

I've gone back and had a careful look over our usage of AttributeProvider and SessionId, and actually it looks like both can be replaced with BacktraceClient's indexer.

IIRC the SessionId was needed since we were setting dying memory before enabling metrics on the BacktraceClient instance (which is the point at which the BacktraceMetrics added its session ID to the client's attributes), so having access to the session ID in advance was needed, but we don't do that any more - it's not much use having the session ID in dying memory anyway if we never got to the point where the Backtrace session metrics were sent anyway.

It also felt cleaner using a member, rather than having to hardcode that we're requesting the "application.session" attribute. Perhaps exposing BacktraceMetrics.ApplicationSessionKey would be sensible?

I can no longer remember why I thought AttributeProvider needed to be exposed - as you mention [here](https://github.com/backtrace-labs/backtrace-unity/pull/233#issuecomment-2582290676), we can use the BacktraceClient indexer directly and it works fine!

konraddysput:
I think we can expose SessionId via metrics - but it would be better to keep attribute provider internal for now.

hymerman:
Sure thing! What's the preferred way to amend a PR here? Should I push another commit with the changes and they'll be squashed here? Sorry, I've not used git or contributed many cls for a long time :D

konraddysput:
Yes - we can squash merge it - feel free to add more commits to each pull request

konraddysput:
hey @hymerman we can also modify your pull requests, if you would like to. We're about to prepare a new backtrace-unity release and I wonder what is the best way to move forward with including your changes in the next release.

hymerman:
Hi Konrad,

Yes please, I think it's best if you modify them - we're also coming up to a release so my focus is 100% on that! Apologies for logging this and then going quiet!

-------- Original Message --------

konraddysput:
We will continue in #238

PR #233, opened 2025-01-10
Expose BacktraceClient.SetAttribute

hymerman:
This is useful to set single attributes without discarding other existing ones, as SetAttributes does. It's useful to set attributes before the BeforeSend callback is called, since then they'll get saved out to be sent with native crashes, where this callback won't get called.

### DIFF
--- a/Runtime/BacktraceClient.cs
+++ b/Runtime/BacktraceClient.cs
@@ -186,6 +186,11 @@ namespace Backtrace.Unity
             }
         }
 
+        public void SetAttribute(string key, string value)
+        {
+            this[key] = value;
+        }
+
         /// <summary>
         /// Number of client attributes
         /// </summary>


konraddysput:
The backtraceClient has `[]` override that allows to set a single attribute. Maybe it's not the clear from our API perspective, so exposing this to set a single attribute makes a lot of sense - however, if you look at the `[]` operator logic, there are more things that are happening (for example: setting an attribute also on the native layer). Can we move the code from the `[]` operator to your method and call your new method in the `[]` operator?

hymerman:
Thanks Konrad - you're absolutely right, the indexer can be called directly, and now I'm no longer sure why I added SetAttribute! I'm happy to either close this, or as you say, move the code around - which would you prefer?

konraddysput:
I think it's fine if we can make it more obvious - I'm happy to add your changes if you change the [] operator implementation and adjust your business logic

konraddysput:
We will continue in #240

PR #235, opened 2025-01-10
Add new error.number attribute to error reports

hymerman:
This starts at 1 and increments with each new report sent. 

This is incredibly useful for several types of query, but particularly:
* Filtering to just the first error in each session, since often subsequent errors are knock-ons from the first
* Sorting errors in a filtered session on the order they happened, which isn't possible when errors happen rapidly enough to get the same timestamps

We have used this change in production on two projects and found it to be very useful.

### DIFF
--- a/Runtime/BacktraceClient.cs
+++ b/Runtime/BacktraceClient.cs
@@ -398,6 +398,13 @@ namespace Backtrace.Unity
         }
 
         /// <summary>
+        /// Count of errors sent so far by this client.
+        /// Added to reports as "error.number" attribute.
+        /// First report will be sent with error.number = 1 (as reports *without* an error.number will appear as the default 0, so we need a way to tell the difference).
+        /// </summary>
+        private int _errorNumber = 1;
+
+        /// <summary>
         /// Initialize new Backtrace integration
         /// </summary>
         /// <param name="configuration">Backtrace configuration scriptable object</param>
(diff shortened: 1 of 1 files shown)


konraddysput:
Hey @hymerman - you can use application.session attribute to get all errors generated in one single session and order them by the timestamp (error creation time). Backtrace can do all of this by you. In your case, if you want to have this attribute, wouldn't it be better to use beforeSend method to add it?

hymerman:
Thanks Konrad - so ordering by timestamp within a session is _nearly_ perfect for that use case, but I did find that sometimes, multiple errors would get the timestamp because of how close together they were.

The main benefit from tracking an error order number though, is in being able to filter to the first error in a session. Quite often, one thing will go wrong, which will put the app in a strange state and then cause many further errors, which normally can't ever happen. Fixing the first error fixes all subsequent errors. So it's usually useful, when dealing with a storm of new errors after a release, to focus first on fixing the errors with "error.number == 1" :)

I logged a support request quite a while back to request this ([here](https://support.backtrace.io/hc/en-us/requests/3246)), but there wasn't any built-in way to do this at the time, which is why I added error.number

konraddysput:
hey @hymerman after checking once again with the team, we would like to give you an option to set this attribute rather than setting it by ourselves. Is there an option for you to use beforeSend callback to include the attribute?

hymerman:
Hi Konrad,

That would work for most use cases, but it's also useful to know for native crashes if there were other errors before them in the session. I could use the session ID (assuming the other things from other PRs have worked, like setting dying memory on console to allow Backtrace to import and match them up!), and in that case it's obvious that the native error was the final one (since it's always a hard crash!), but it also feels tidier if the native error has the same error.number attribute that the other errors have, so it can be viewed the same way in the web UI.

That's just my opinion though - I appreciate that it's a trade-off against complexity, so of course it's up to you :)

Thanks,
Ben

konraddysput:
Please keep in mind you can use your SetAttribute method that will set error.count value also on native!

PR #236, opened 2025-01-10
Handle missing Backtrace/Crashpad DLLs gracefully

hymerman:
Log a warning instead of allowing an unhandled exception to escape, if the Backtrace/Crashpad DLLs can't be found, or loaded.

This can happen if e.g. the system doesn't have the version of the VC runtime installed which is required to use these DLLs.

### DIFF
--- a/Runtime/Native/Windows/NativeClient.cs
+++ b/Runtime/Native/Windows/NativeClient.cs
@@ -108,6 +108,13 @@ namespace Backtrace.Unity.Runtime.Native.Windows
                 return;
             }
 
+            var backtraceCrashpadHandlerPath = GetDefaultPathToBacktraceCrashpadHandler(pluginDirectoryPath);
+            if (string.IsNullOrEmpty(backtraceCrashpadHandlerPath) || !File.Exists(backtraceCrashpadHandlerPath))
+            {
+                Debug.LogWarning("Backtrace native integration status: Cannot find path to Backtrace Crashpad handler.");
+                return;
+            }
+
             var databasePath = _configuration.CrashpadDatabasePath;
             if (string.IsNullOrEmpty(databasePath) || !Directory.Exists(_configuration.GetFullDatabasePath()))
             {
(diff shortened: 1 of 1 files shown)


Runtime/Native/Windows/NativeClient.cs
```diff
@@ -122,12 +129,19 @@ private void HandleNativeCrashes(IDictionary<string, string> clientAttributes, I
...
+            try
+            {
+                CaptureNativeCrashes = Initialize(
+                    minidumpUrl,
+                    databasePath,
+                    crashpadHandlerPath,
+                    attachments.ToArray(),
+                    attachments.Count() );
+            }
+            catch ( DllNotFoundException )
```
konraddysput, line 141: In this case it would be better if we fail fast and hard. This situation most likely won't happen in production, because you can easily spot it during the development/release preparation cycle. In the end if we hide a huge issue, we can have a lot of situation where people could lose crashes.

konraddysput (changes requested):

Runtime/Native/Windows/NativeClient.cs
hymerman: Ah, I should have added - this issue happened a lot in production, since 'in the wild', customers often had outdated versions of the VC runtimes installed. These are customers installing our games through Steam, which *does* install VC runtimes for them automatically, but the exact versions are often quite outdated. I spoke with our technical contacts at Valve and they mentioned that they don't make much of an effort to keep these up to date, and they didn't update the runtime versions after we requested them. The same issue happened with NVidia Ansel DLLs that we shipped with the game.

We can ship the game with extra install steps and bundle the correct VC runtime versions, but this is considered 'advanced' setup for Steam apps, and I imagine the vast majority of developers don't do this. Considering how hard it was to track this issue down, I think in this case it may be best to be permissive (I'm usually a huge fan of failing fast and hard!).

Perhaps it could instead be logged as an error (which then gets sent to Backtrace)? So it can be explicitly flagged to tell developers that they need to do some more installation setup? Also the Backtrace developer documentation would need to mention that a particular VC runtime version would need to be distributed and installed with the application.

konraddysput: ahh ok - this is super interesting insight - thank you for sharing it! I will double check it with the team to make sure we're all on the same page and share more info with you on this pull request. thank you!

perf2711 (approved):
Looks good to me.

konraddysput (approved):

PR #240, opened 2025-01-24
Feature/set attribute

konraddysput:
# Why

Backtrace Unity plugin has `[]` operator to set a key-value pair that represents an attribute. However, our users don't know about it, that is why we are preparing another option to set a key-value pair in the client. 

The `SetAttributes` method was also extended. Now every time, when attributes are set, they're also added to the native layer.

### DIFF
--- a/Runtime/BacktraceClient.cs
+++ b/Runtime/BacktraceClient.cs
@@ -142,11 +142,7 @@ namespace Backtrace.Unity
             }
             set
             {
-                AttributeProvider[index] = value;
-                if (_nativeClient != null)
-                {
-                    _nativeClient.SetAttribute(index, value);
-                }
+                SetAttribute(index, value);
             }
         }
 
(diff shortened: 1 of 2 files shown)


Runtime/BacktraceClient.cs
```diff
@@ -170,6 +166,27 @@ public IEnumerable<string> GetAttachments()
...
+        /// </summary>
+        /// <param name="key">Attribute name</param>
+        /// <param name="value">Attribute value</param>
+        /// <returns>True, if the value was added. Otherwise false.</returns>
+        public bool SetAttribute(string key, string value) 
+        {
+            if (string.IsNullOrEmpty(key)) 
+            {
+                return false;
+            }
```
perf2711, line 180: Why not throw an exception here?

konraddysput: I prefer to stay with the solution where we won't throw an error in case of a user mistake.

perf2711 (approved):

PR #243, opened 2025-03-04
Make env variable optional

konraddysput:
# Why

Environment variables are hard to manage for our users. They can contain PII info/user secrets that are hard to remove on the SDK initialization. To minimalize the risk of submitting something unwanted, this pull request disabled by default env variables and the user (if it needs to) can include them.

ref: INT-80

### DIFF
--- a/Runtime/Model/JsonData/Annotations.cs
+++ b/Runtime/Model/JsonData/Annotations.cs
@@ -20,27 +20,17 @@ namespace Backtrace.Unity.Model.JsonData
         internal static Dictionary<string, string> _environmentVariablesCache;
 
         /// <summary>
-        /// Determinate if static helper should load environment variables or not.
-        /// </summary>
-        internal static bool VariablesLoaded;
-
-        /// <summary>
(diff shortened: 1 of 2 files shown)


Runtime/Model/JsonData/Annotations.cs
```diff
@@ -19,28 +19,18 @@ public class Annotations
...
-                {
-                    _environmentVariablesCache = SetEnvironmentVariables();
-                    VariablesLoaded = true;
-                }
                 return _environmentVariablesCache;
             }
             set
             {
-                _environmentVariablesCache = value;
+                _environmentVariablesCache = SetEnvironmentVariables(_environmentVariablesCache);
```
perf2711, line 33: Did you mean 
```suggestion
                _environmentVariablesCache = SetEnvironmentVariables(value);
```
?

konraddysput: You're correct. I cherry-picked not enough changes. thanks!

perf2711 (approved):

PR #245, opened 2025-04-07
Generate valid machine id 

konraddysput:
# Why

In the past, we used Unity ID to generate a user GUID attribute. If the Unity identifier was unavailable for any reason, we fall back to the network identifier. If both methods failed, we generated a random guid. Later, we also introduced storing the guid in the session memory. 

Recently we noticed, the user identifier provided by Unity is not a valid guid. We thought it is based on the quick tests and results in Backtrace, but it looks like it's not on Windows at least. 

To mitigate the issue:
- When a user identifier is read, we verify if its a valid uuid. If not, we're converting a string into a uuid representation. 
- The code was changed to verify potential tests cases and strategies. 

Testing strategy:
- unity tests
- manual tests

Expected results: 

- if the guid is invalid, always generate a valid guid based on the string we stored in memory or we got from Unity,\

ref: INT-93


Runtime/Common/GuidHelper.cs
```diff
@@ -23,5 +25,20 @@ public static bool IsNullOrEmpty(string guid)
             const string emptyGuid = "00000000-0000-0000-0000-000000000000";
             return string.IsNullOrEmpty(guid) || guid == emptyGuid;
         }
+
+        /// <summary>
+        /// Converts a random string into a guid representation.
+        /// </summary>
+        public static Guid FromString(string value)
```
perf2711, line 32: Can we rename this to `FromMD5String`? `FromString` suggests that a GUID string can be provided here.

Runtime/Model/DataProvider/IMachineIdentifierDataProvider.cs
```diff
@@ -0,0 +1,8 @@
+﻿[assembly: System.Runtime.CompilerServices.InternalsVisibleTo("Backtrace.Unity.Tests.Runtime")]
+namespace Backtrace.Unity.Model.DataProvider
+{
+    internal interface IMachineIdentifierDataProvider
```
perf2711, line 4: This only provides the machine identifier. Why not name this `IMachineIdentifierProvider`?

Runtime/Model/DataProvider/UnityMachineIdentifierProvider.cs
```diff
@@ -0,0 +1,37 @@
...
+            if (!IsValidIdentifier())
+            {
+                return null;
+            }
+
+            if (Guid.TryParse(_deviceUniqueIdentifier, out Guid unityUuidGuid))
+            {
+                return unityUuidGuid.ToString();
+            }
+            return GuidHelper.FromString(_deviceUniqueIdentifier).ToString();
```
perf2711, line 29: How often the ID will be generated? Maybe it's a good idea to generate this in the constructor, or cache the value somewhere?

perf2711: Oh, I see that in `MachineIdStorage` it is retrieved only once? If so, feel free to discard this comment.

perf2711 (approved):

Runtime/Common/GuidHelper.cs
konraddysput: MD5 is an algorithm behind the scene that we're using to convert string into bytes that later we can use to generate GUID. It doesn't make sense to include MD5 in the name in my opinion

Runtime/Model/DataProvider/IMachineIdentifierDataProvider.cs
konraddysput: done

Runtime/Model/DataProvider/UnityMachineIdentifierProvider.cs
konraddysput: to confirm - yes. we're doing it only once.

PR #250, opened 2025-05-21
Updated PLCrashReporter static dependencies for Device & Simulator

melekr:
This PR adds PLCrashReporter static libraries to support native crash reporting in iOS Unity builds across device and simulator targets.

- Adds libBacktrace-Unity-Cocoa-Sim.a for [ arm64, x86_64] simulators.
- Adds libBacktrace-Unity-Cocoa-Sim.a for [ arm64, arm64v7] devices.

ref: BT-5766


melekr:
generated with the 2022 LTS, I’ll trim the .meta back to a minimal scheme for better Unity 5 compatibility / less warnings.

PR #278, opened 2026-03-05
Fix crash when parsing native stack frames without symbols

BartoszLitwiniuk:
## Summary
Fixes a crash in the native stack trace parser when frames are missing symbol information. Refactors BacktraceRawStackTraceParser for safer parsing and adds unit test coverage.

It's a follow up of changes made in https://github.com/backtrace-labs/backtrace-unity/pull/258 and bug `BT-5953`.

## Changes
- Added bounds checks in `SetNativeStackTraceInformation` to prevent out-of-range access when a native frame lacks symbol data
- Extract parsing logic to `BacktraceRawStackTraceParser` for clearer, more defensive frame parsing
- Added debug logging to aid future diagnosis of malformed frames
- Added `BacktraceRawStackTraceParserTests` covering native frame parsing edge cases
- Use `BacktraceRawStackTraceParser` in `BacktraceUnhandledException` instead of mixing unhandled exception handler with parsing logic

## Testing
- New unit tests validate safe parsing of native frames, including frames without symbols
- Existing tests continue to pass

[BT-5953]: https://saucedev.atlassian.net/browse/BT-5953?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


melekr (approved):

melekr:
can you link sample BT reports @BartoszLitwiniuk ?

.gitignore
```diff
@@ -67,3 +67,7 @@ sysinfo.txt
 
 # Crashlytics generated file
 crashlytics-build.properties
+
+# Platforms
+iOS/
+Mac/
```
melekr: what's the reason here?

BartoszLitwiniuk: reverted

Runtime/Model/BacktraceRawStackTraceParser.cs
```diff
@@ -0,0 +1,474 @@
...
+
+        private BacktraceStackFrame TryParseFrameOrFallback(string frame)
+        {
+            try
+            {
+                string frameString = frame.Trim();
+
+                // validate if stack trace has exception header               
+                int methodNameEndIndex = frameString.IndexOf(')');
+                int openParentIndex = frameString.LastIndexOf('(', methodNameEndIndex); // we require a '(' that appears before this ')'
```
melekr, line 51: This can throw before the invalid-frame check. 

If `frameString` is non-empty and has no `)`, then `methodNameEndIndex` is `-1`,
and `LastIndexOf('(', -1)` throws `ArgumentOutOfRangeException`.

That changes behaviour from the previous implementation which preserved frames without `)`as raw `BacktraceStackFrame` instances.

With the current code, the `catch` returns `null`, and `ConvertStackFrames` drops the frame entirely.

I suggest something like this : 

```suggestion
int methodNameEndIndex = frameString.IndexOf(')');
if (methodNameEndIndex == -1)
{
    return new BacktraceStackFrame
    {
        FunctionName = frame,
        InvalidFrame = true
    };
}

int openParentIndex = frameString.LastIndexOf('(', methodNameEndIndex);
if (openParentIndex == -1 || openParentIndex > methodNameEndIndex)
{
    return new BacktraceStackFrame
    {
        FunctionName = frame,
        InvalidFrame = true
    };
}
```

Runtime/Model/BacktraceRawStackTraceParser.cs
```diff
@@ -0,0 +1,474 @@
...
+                    Debug.LogWarning($"Invalid stack frame format: '{frameString}'.");
+                    return new BacktraceStackFrame { FunctionName = frame };
+                }
+
+                return ParseStacktraceFrame(frameString, methodNameEndIndex);
+            }
+            catch (Exception e)
+            {
+                Debug.LogError($"Exception while parsing stack frame: '{frame}'. Exception: {e}");
+                return null;
```
melekr: Returning `null` here causes the frame to be dropped. 

I think parser failure should preserve the raw frame, not remove diagnostic data from the report.

Also, logging with `Debug.LogError` from report construction is risky in this SDK: Unity error logs are reportable, so a parser failure can create a secondary Backtrace report or another capture.

Runtime/Model/BacktraceRawStackTraceParser.cs
```diff
@@ -0,0 +1,474 @@
...
+                string frameString = frame.Trim();
+
+                // validate if stack trace has exception header               
+                int methodNameEndIndex = frameString.IndexOf(')');
+                int openParentIndex = frameString.LastIndexOf('(', methodNameEndIndex); // we require a '(' that appears before this ')'
+
+                if (methodNameEndIndex == -1 || openParentIndex == -1 || openParentIndex > methodNameEndIndex)
+                {
+                    // If either index is missing, it's an invalid frame
+                    Debug.LogWarning($"Invalid stack frame format: '{frameString}'.");
```
melekr: I would avoid logging from the parser for the same reason.

Runtime/Model/BacktraceRawStackTraceParser.cs
```diff
@@ -0,0 +1,474 @@
...
+            };
+            if (!frameString.StartsWith("#", StringComparison.Ordinal))
+            {
+                //handle situation when we detected jit stack trace
+                // but jit stack trace doesn't start with #
+                stackFrame.FunctionName = frameString;
+                return stackFrame;
+            }
+
+            frameString = frameString.Substring(frameString.IndexOf(' ')).Trim();
```
melekr: This remains unsafe;

If a JIT frame starts with `#` but does not contain a `space`, `IndexOf(' ')` returns `-1`  and `Substring(-1)` throws. 

With the current wrapper logic, that exception causes the frame to be dropped.

melekr: something like :

```suggestion
            int firstSpace = frameString.IndexOf(' ');
if (firstSpace == -1 || firstSpace == frameString.Length - 1)
{
    stackFrame.FunctionName = frameString;
    return stackFrame;
}

frameString = frameString.Substring(firstSpace + 1).Trim();
```

Tests/Runtime/Model/BacktraceRawStackTraceParserTests.cs
```diff
@@ -0,0 +1,35 @@
+using Backtrace.Unity.Model;
+using Backtrace.Unity.Types;
+using NUnit.Framework;
+
+namespace Backtrace.Unity.Tests.Runtime
+{
+    public class BacktraceRawStackTraceParserTests
```
melekr, line 7: needs more coverage especially through `ConvertStackFrames` and `ConvertStackFrames`

Runtime/Types/BacktraceStackFrameType.cs
```diff
@@ -1,6 +1,6 @@
 ﻿namespace Backtrace.Unity.Types
 {
-    enum BacktraceStackFrameType
+    public enum BacktraceStackFrameType
```
melekr: Why does this need to become public ?

Tests/Runtime/Model/BacktraceRawStackTraceParserTests.cs
melekr, line 7: Let's add an end-to-end serialization test or attach a sample Backtrace report here.

melekr (changes requested):

Runtime/Types/BacktraceStackFrameType.cs
BartoszLitwiniuk: Fixed - I've changed a bit tests and reverted that.

Runtime/Model/BacktraceRawStackTraceParser.cs
BartoszLitwiniuk: Remove logging here.

Runtime/Model/BacktraceRawStackTraceParser.cs
BartoszLitwiniuk: Log removed and now I'm returning raw frame

Runtime/Model/BacktraceRawStackTraceParser.cs
BartoszLitwiniuk: Agree, fixed

Runtime/Model/BacktraceRawStackTraceParser.cs
BartoszLitwiniuk: true, fixed

BartoszLitwiniuk: good point and solid fix - applied

Tests/Runtime/Model/BacktraceRawStackTraceParserTests.cs
BartoszLitwiniuk: There are already end-to-end tests here https://github.com/backtrace-labs/backtrace-unity/blob/master/Tests/Runtime/BacktraceStackTraceTests.cs and all of them passed

BartoszLitwiniuk: There are already end-to-end tests here https://github.com/backtrace-labs/backtrace-unity/blob/master/Tests/Runtime/BacktraceStackTraceTests.cs and all of them passed

Runtime/Model/BacktraceRawStackTraceParser.cs
```diff
@@ -0,0 +1,484 @@
...
+                BacktraceStackFrame convertedFrame = TryParseFrameOrFallback(frame);
+                if (convertedFrame != null)
+                {
+                    result.Add(convertedFrame);
+                }
+            }
+            return result;
+        }
+
+        private BacktraceStackFrame TryParseFrameOrFallback(string frame)
```
melekr, line 37: how about : 


```suggestion
        private BacktraceStackFrame TryParseFrameOrFallback(string frame)
{
    try
    {
        string frameString = frame.Trim();

        if (frameString.StartsWith("0x", StringComparison.Ordinal))
        {
            return ParseNativeFrame(frameString);
        }

        if (frameString.StartsWith("#", StringComparison.Ordinal))
        {
            return SetJITStackTraceInformation(frameString);
        }

        int methodNameEndIndex = frameString.IndexOf(')');
        if (methodNameEndIndex == -1)
        {
            return new BacktraceStackFrame
            {
                FunctionName = frame,
                InvalidFrame = true
            };
        }

        int openParentIndex = frameString.LastIndexOf('(', methodNameEndIndex);
        if (openParentIndex == -1 || openParentIndex > methodNameEndIndex)
        {
            return new BacktraceStackFrame
            {
                FunctionName = frame,
                InvalidFrame = true
            };
        }

        return ParseStacktraceFrame(frameString, methodNameEndIndex);
    }
    catch
    {
        return new BacktraceStackFrame
        {
            FunctionName = frame,
            InvalidFrame = true
        };
    }
}
```

melekr (approved):
lgtm.

let's add the tests as discussed.

melekr (approved):

PR #288, opened 2026-05-11
release JNI refs + detach current thread

HamzaRino:
1- JNI refs leak, hard limit of 512 entries for old Android versions.

`AndroidJNI.NewStringUTF()` allocates a JNI local reference. 
The old code passed them inline:
`AddAttribute(AndroidJNI.NewStringUTF(key), AndroidJNI.NewStringUTF(value)); // refs leaked`

Local refs are never freed this way. The JNI local ref table has a hard limit (~512 entries on most Android VMs).
Calling this in the `foreach` loop over `backtraceAttributes` with many entries would eventually crash with a `JNI ERROR: local reference table overflow`.
The fix: store the `IntPtr`, call `DeleteLocalRef` in `finally` is the correct pattern.

Arrays created via `AndroidJNIHelper.ConvertToJNIArray` are JNI local object array refs, `DeleteLocalRef` on them is correct. They're now also cleaned up in the finally.

2- thread starvation

The old ANR thread code:

```
if (AndroidJNI.AttachCurrentThread() == 0)
{
    AddAttribute(...);
    NativeReport(...);
    AddAttribute(...);
    // no DetachCurrentThread!
}
```

Every time an ANR was detected, the background thread was attached to the JVM and never detached.
This leaks a JVM thread reference on every ANR occurrence.
The fix wraps the JNI work in `try/finally` with `DetachCurrentThread()` in the outer finally correct.

 note: changes suggested by gemini CLI & reviewed by Claude Code.

### DIFF
--- a/Runtime/Native/Android/NativeClient.cs
+++ b/Runtime/Native/Android/NativeClient.cs
@@ -376,8 +376,18 @@ namespace Backtrace.Unity.Runtime.Native.Android
             }
 
             foreach (var attribute in backtraceAttributes)
-            {   
-                AddAttribute(AndroidJNI.NewStringUTF(attribute.Key), AndroidJNI.NewStringUTF(attribute.Value));
+            {
+                var keyPtr = AndroidJNI.NewStringUTF(attribute.Key);
+                var valuePtr = AndroidJNI.NewStringUTF(attribute.Value);
+                try
+                {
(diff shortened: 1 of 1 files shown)


JohnTube:
@melekr can you please explain why you closed this MR?

PR #294, opened 2026-07-02
chore: fix failing unity tests on windows player builds

KishanPRao:
## Summary

Fixes tests that passed in the Editor but failed with Player, fixed a bug in the Windows native client which was causing mishandling of legacy attributes, guard a few tests that should only run in the Editor or Android.

## Changes

- `NativeClient`: `GetScopedAttributes` no longer returns early when there's no scoped list, so legacy attributes are still migrated on upgrade. Save PlayerPrefs after removing legacy keys.
- Renamed `HasScopedAttributesEmpty` to `HasScopedAttributes`, which returns true when attributes exist.
- `ScopedNativeAttributesTests`: clear PlayerPrefs in `[SetUp]` so tests don't inherit state from earlier Player runs.
- `BacktraceClientDisableTests`: skip the `DisableInEditor` tests in Player builds, should be Editor only.
- `BacktraceStackTraceTests`: skip the Android stack trace tests in Player builds, should be Android/Editor only

### DIFF
--- a/Runtime/Native/Windows/NativeClient.cs
+++ b/Runtime/Native/Windows/NativeClient.cs
@@ -424,6 +424,7 @@ namespace Backtrace.Unity.Runtime.Native.Windows
                         { SessionKey, BacktraceMetrics.ApplicationSessionKey }
                     };
 
+            bool removedLegacy = false;
             foreach (var legacyAttribute in legacyAttributes)
             {
                 string legacyAttributeValue = PlayerPrefs.GetString(legacyAttribute.Key, string.Empty);
@@ -431,7 +432,13 @@ namespace Backtrace.Unity.Runtime.Native.Windows
                 {
                     PlayerPrefs.DeleteKey(legacyAttribute.Key);
                     result[legacyAttribute.Value] = legacyAttributeValue;
+                    removedLegacy = true;
                 }
+            }
+
+            if (removedLegacy)
+            {
+                PlayerPrefs.Save();
             }
             return result;
         }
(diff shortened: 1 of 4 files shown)


melekr (approved):
Nice cleanup. The platform specific skips look appropriate: Android stack trace parsing is compiled only for Android/Editor and the `DisableInEditor` assertions are editor only by design.

One small suggestion: `GetScopedAttributes()`now reads legacy PlayerPrefs attributes even when the scoped key list is missing, but the legacy migration path deletes those old PlayerPrefs keys without calling `PlayerPrefs.Save()`. Since this is part of the Windows startup/minidump flow, an explicit save after deleting legacy keys would make the cleanup deterministic and avoid relying on a later PlayerPrefs flush.

Maybe worth adding `PlayerPrefs.Save()` at the end of `ScopedNativeAttributesTests.CleanState()` after `PlayerPrefs.DeleteAll()` in Player test runs?

KishanPRao:
> Nice cleanup. The platform specific skips look appropriate: Android stack trace parsing is compiled only for Android/Editor and the `DisableInEditor` assertions are editor only by design.
> 
> One small suggestion: `GetScopedAttributes()`now reads legacy PlayerPrefs attributes even when the scoped key list is missing, but the legacy migration path deletes those old PlayerPrefs keys without calling `PlayerPrefs.Save()`. Since this is part of the Windows startup/minidump flow, an explicit save after deleting legacy keys would make the cleanup deterministic and avoid relying on a later PlayerPrefs flush.
> 
> Maybe worth adding `PlayerPrefs.Save()` at the end of `ScopedNativeAttributesTests.CleanState()` after `PlayerPrefs.DeleteAll()` in Player test runs?

Good suggestions!
I've made changes to perform a `Save` on the prefs, if legacy attributes were processed & deleted from prefs.

For the Player test runs, since we run `CleanState` before every test, the in-memory store should always be empty. `Save` would write into disk, but this would not be that useful here, since we always read from a clean state, and it'd just add unused I/O operations. I'd opt to avoid adding it.

PR #296, opened 2026-07-09
bugfix: fix potential thread-safety issue in BacktraceLogManager

KishanPRao:
## Summary

`ToSourceCode()` reads the log queue without the lock that `Enqueue()`/`Dequeue()` use, so a concurrent log from a background thread could corrupt the read. This makes the manager thread-safe on its own instead of relying on the caller to lock.

## Changes

- `BacktraceLogManager`: Wrap the `LogQueue` read in `ToSourceCode()` with `lockObject`.

ref: BT-7318

### DIFF
--- a/Runtime/Model/BacktraceLogManager.cs
+++ b/Runtime/Model/BacktraceLogManager.cs
@@ -101,17 +101,18 @@ namespace Backtrace.Unity.Model
         /// <returns>Source code</returns>
         public string ToSourceCode()
         {
+            string[] logs;
             lock (lockObject)
             {
-                var stringBuilder = new StringBuilder();
-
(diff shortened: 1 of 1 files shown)


melekr (approved):
Non-blocking: the lock scope could be narrowed to only the `ToArray()` copy, then build the `StringBuilder` after releasing the lock. Still looks good given the small default log count, but the narrower scope would be cleaner.

KishanPRao:
> Non-blocking: the lock scope could be narrowed to only the `ToArray()` copy, then build the `StringBuilder` after releasing the lock. Still looks good given the small default log count, but the narrower scope would be cleaner.

Good idea, I've narrowed the lock's scope.

